PR #2, opened 2016-05-22
Updating vim colors on Vim, MacVim and NeoVim

lucazz:
Updating vim colors so Vim and NeoVim so `Specialkeys` and `Nontext` gets rendered the same way on ViM, NeoViM and MacViM

Before: 

After:


lucazz:
Here's what the VertSplit fix look like:

Before: before

After: after

PR #4, opened 2016-05-24
Reintroduce LineNr and CursorLine GUI colors

niclas-ahden:
GUI colors for line numbers and the current line highlight was removed in c661a57. It seems like a mistake as the changes seem to originate from zenorocha/dracula-theme/issues/80 which is about terminal colors, not GUI. This PR reintroduces the removed GUI colors.

Before:
image

After:
image

PR #5, opened 2016-05-25
Align colors with the zenorocha/dracula-theme palette

niclas-ahden:
The main repo [zenorocha/dracula-theme](https://github.com/zenorocha/dracula-theme) has a color palette which is the source of truth for the theme. The vim theme hasn't been keeping up with it and this PR aims to fix that. Follow-up from #4.

Before:
image
After:
image

PS. I personally prefer the Before-picture since the darker background makes it more readable and the colors pop more (click the images to see the difference). That's a separate issue related to the color palette though.

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -22,10 +22,10 @@ let g:colors_name = "dracula"
 
 hi Cursor ctermfg=17 ctermbg=231 cterm=NONE guifg=#282a36 guibg=#f8f8f0 gui=NONE
 hi Visual ctermfg=NONE ctermbg=236 cterm=NONE guifg=NONE guibg=#44475a gui=NONE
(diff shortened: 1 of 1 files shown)


nesl247:
There's still a few more instances in the config of colors not in our color palette. I'm not a VIM expert though to know whether or not it's because we don't have a good color to fit in those spots or not. Can you take a look again?

niclas-ahden:
Yes, there are some colors for which there are no good options in the palette (for example `LineNr` and the Vim-specific `SpecialKeys`). I think this is a good enough start to get us closer to the palette.

nesl247:
We should probably discuss getting a more complete palette then so other editors don't have this same issue. We can do that in the main repo though.

PR #6, opened 2016-05-25
Fix contributors link in README

niclas-ahden:
It was pointing to Alfred, but this is Vim :v:


nesl247:
Thanks for catching that. We have 33 repositories if I recall, and when you do the same thing so many times in a row, especially late at night, small things slip by.

If https://xkcd.com/1319/ wasn't true I'd have automated the changes.

niclas-ahden:
Happy to help, thanks for setting up all the repos! 👍

PR #16, opened 2016-09-21
Add better support for Elixir and VimScript

dkarter:
Also included:
- specify colors for Type which is used a lot for C and many other
languages.
- specify better colors for CursorColumn which was broken before

## Screenshots for VimScript and CursorColumn:

BEFORE
image

AFTER
image

## Screenshots for Elixir:

BEFORE
image

AFTER
image


heinst:
👍 Thanks

PR #18, opened 2016-10-01
make pmenu more readable

sdemura:
Adjust pmenu colors to make them more readable.

Before:
before

After:
after


heinst:
Thanks!

PR #22, opened 2016-11-22
Indicate in airline that the current buffer has unsaved changes

simplay:
Color the airline statusbar in gray when there are unsaved changes (all modes).

patched_airline

**Before**:
before

**After**:
after


simplay:
ping @heinst / @zenorocha

PR #30, opened 2017-05-05
Set guifg for rubyInstanceVariable

hzchirs:
There's no highlight for rubyInstanceVariable in neovim.
Referenced to other editor's syntax highlight and add #ffb86c to rubyInstanceVariable guifg.

Before:

After:


hzchirs:
@heinst are there any problems on my pull request?

PR #38, opened 2017-07-12
fix typo foldcolumn

turnon:
the foldcolumn color cofig doesn't take effect because of typo (env: centos7, vim7.4)

above: before
below: after

before

after


dsifford:
Thanks @turnon 

Sorry for the delay.

PR #43, opened 2017-09-04
correct tab view

solarmerry:
before / after
bf af


BlakeWilliams:
This looks good, thanks!

PR #53, opened 2018-01-10
Add SpecialComment highlight

topecongiro:
Highlight `SpecialComment` in the same way as `Comment`.

#### Before
screenshot_20180110_121500

#### After
screenshot_20180110_121337


benanders:
I was just about to submit a pull request for this, but you beat me to it by only 11 hours! Could we please add this? Maybe there should be some discussion regarding whether doc comments should be made to stand out more than regular comments (rather than making them the same color)? Maybe make doc comments a slightly lighter version of the purple comment color?

BlakeWilliams:
> Maybe there should be some discussion regarding whether doc comments should be made to stand out more than regular comments (rather than making them the same color)? Maybe make doc comments a slightly lighter version of the purple comment color?

Not a bad idea, maybe italicizing the text could be another alternative? In the meantime I'm merging this comment since it's better than nothing.

Thanks for the contribution @topecongiro!

PR #71, opened 2018-04-05
add support for plasticboy/vim-markdown syntax

dsifford:
This is the first of many planned syntax extensions to provide better syntax highlighting support in accordance to the running spec.

Because this syntax uses a lot of the built-in html highlight groups, it needed to be done as a ftplugin so that the highlight tweaks only apply for markdown.

This begs the question: Should we be doing most/all of the syntax highlighting as ftplugins? Would that offer better performance?

Also -- @zenorocha Can we add @benknoble as a maintainer here? And if possible, can we remove the other two listed maintainers? I tried reaching out to both of them twice now with no reply. I think it's safe to say they've lost interest.

Thanks :+1:

### DIFF
--- a/ftplugin/markdown.vim
+++ b/ftplugin/markdown.vim
@@ -1,0 +1,32 @@
+
+
+" plasticboy/vim-markdown {{{1
+
+hi! link htmlH1 DraculaPurpleBold
+
+hi! link htmlItalic DraculaYellowItalic
+hi! link mkdItalic DraculaYellowItalic
+
+hi! link htmlBold DraculaOrangeBold
+hi! link mkdBold DraculaOrangeBold
+
+hi! link htmlBoldItalic DraculaOrangeBoldItalic
+hi! link mkdBoldItalic DraculaOrangeBoldItalic
+
+hi! link mkdRule DraculaComment
+
+hi! link mkdBlockquote DraculaYellowItalic
+
+hi! link mkdListItem DraculaCyan
+
+hi! link mkdCode DraculaGreen
+hi! link mkdCodeStart DraculaGreen
+hi! link mkdCodeEnd DraculaGreen
+
+hi! link mkdLink DraculaPink
+hi! link mkdUrl DraculaLink
+hi! link mkdInlineUrl DraculaLink
+
+"}}}1
+
+" vim: fdm=marker ts=2 sts=2 sw=2:
--- a/ftplugin/markdown.vim
+++ b/ftplugin/markdown.vim
@@ -1,5 +1,3 @@
-
-
 " plasticboy/vim-markdown {{{1
 
 hi! link htmlH1 DraculaPurpleBold
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -386,7 +386,7 @@ hi! link markdownBoldItalic DraculaOrangeBoldItalic
 
 hi! link markdownBlockquote DraculaCyan
 
-hi! link markdownCode DraculaGreenItalic
+hi! link markdownCode DraculaGreen
 hi! link markdownCodeDelimiter DraculaGreen
 
 hi! link markdownListMarker DraculaCyan


benknoble:
I was just about to start a review and note that

1. I don't understand what the reference to `plasticboy` is, and
2. What are the highlight groups `mkd[...]`, how are they different from `markdown[...]` that are already part of the color scheme

Then I realized this is probably a plugin. I don't know anything about it, since I don't use it.

ftplugin/markdown.vim
```diff
@@ -0,0 +1,32 @@
+
+
```
benknoble: Blank lines to serve a purpose? Minor aesthetics, but still.

ftplugin/markdown.vim
```diff
@@ -0,0 +1,32 @@
...
+hi! link htmlBold DraculaOrangeBold
+hi! link mkdBold DraculaOrangeBold
+
+hi! link htmlBoldItalic DraculaOrangeBoldItalic
+hi! link mkdBoldItalic DraculaOrangeBoldItalic
+
+hi! link mkdRule DraculaComment
+
+hi! link mkdBlockquote DraculaYellowItalic
```
benknoble: In standard dracula, this is `DraculaCyan`

ftplugin/markdown.vim
```diff
@@ -0,0 +1,32 @@
...
+
+hi! link mkdListItem DraculaCyan
+
+hi! link mkdCode DraculaGreen
```
benknoble: In standard dracula, this is `DraculaGreenItalic`

dsifford:
@benknoble Thanks for the review!

Haha, yes [plasticboy/vim-markdown](https://github.com/plasticboy/vim-markdown) is indeed a plugin. 

I know supporting plugins can potentially create a slippery precedent for users wanting other plugins that they use supported, but I think if we approach it carefully, it can be manageable.

In this case, `plasticboy/vim-markdown` has 2400 stars on GitHub and is actively maintained, so I figured it would be something that large chunk of our users would find useful. I use this plugin exclusively for the improved syntax. I have none of the other features enabled.

Other plugins that I hope to add support for are [leafgarland/typescript-vim](https://github.com/leafgarland/typescript-vim) and [pangloss/vim-javascript](https://github.com/pangloss/vim-javascript) (just to name a few), both of which have a much better base syntax and a pretty good looking coverage of the latest language features.

Initially I was going to only add in the highlight links for these types of syntaxes if the user has the syntax installed, but I'm not aware of a reliable enough way to do that (suggestions?).

ftplugin/markdown.vim
dsifford: Nope. Good catch!

ftplugin/markdown.vim
dsifford: Nope. Good catch!

ftplugin/markdown.vim
dsifford: Yeah, so both are technically incorrect as it should be cyan for the `>` and yellow italic for the actual text.

The built-in syntax only targets the `>` so that's why it's cyan.

This one only targets everything, with no distinction of the `>`, so I opted to make that yellow italic. Thoughts?

ftplugin/markdown.vim
dsifford: Hm... Not sure why I made the base syntax italic. The spec says it should be non-italic. 

I'll fix the base implementation. Really good catch!

dsifford:
Yeah, definitely tricky...

Just to clarify...

> b:current_syntax is set to mkd (the standard syntax files set markdown)

Current syntax will still be markdown... `mcd` is just a special name that they gave to the syntax groups, so that wouldn't work.

I was thinking of maybe checking the `:scriptnames` path or something like that, but it would be fragile.

Another way would be to use, for example, `g:plugs` for users with `vim-plugged`, which would absolutely work, but there's too many different plugin managers to set up a reliable way of checking these types of variables and not all of them (I would assume) have that capability.

---

But anyway, we good to merge?

Also, I'll reach out to @zenorocha by email if I don't hear back Re: adding you to this repo as a maintainer. 👍

PR #74, opened 2018-04-19
scss adjustments

dsifford:
Added a few small adjustments for to scss for spec alignment

cc: @benknoble

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -352,10 +352,12 @@ hi! link cssBraces Delimiter
 hi! link cssSelectorOp Delimiter
 hi! link cssFunctionComma Delimiter
 hi! link cssAttributeSelector DraculaGreenItalic
+hi! link cssVendor DraculaGreenItalic
 hi! link cssAttrRegion DraculaPink
 hi! link cssUnitDecorators DraculaPink
 hi! link cssProp DraculaCyan
 hi! link cssPseudoClassId DraculaGreenItalic
+hi! link cssNoise DraculaPink
 
 "}}}2
 " Git Commit: {{{2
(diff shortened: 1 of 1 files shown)


colors/dracula.vim
```diff
@@ -431,15 +433,16 @@ hi! link rubyGlobalVariable DraculaPurple
...
-hi! link yamlAnchor DraculaPinkItalic
-hi! link yamlAlias DraculaGreenItalicUnderline
-hi! link yamlNodeTag DraculaPink
-hi! link yamlFlowCollection DraculaPink
-hi! link yamlFlowIndicator Delimiter
+hi! link sassProperty cssProp
+hi! link sassMixinName Function
+hi! link sassClass DraculaGreen
+hi! link sassClassChar DraculaGreen
+hi! link sassInterpolationDelimiter DraculaPink
```
benknoble: Is there a reason not to link to the `Delimiter` group? Although it appears to be a different color...

colors/dracula.vim
```diff
@@ -431,15 +433,16 @@ hi! link rubyGlobalVariable DraculaPurple
...
+hi! link sassVariableAssignment DraculaPink
+hi! link sassId DraculaGreen
```
benknoble: Similarly, what about the `Identifier` group? Also appears to be a different color...

colors/dracula.vim
dsifford: Not sure I fully understand the question here. I'm assuming you're asking why we don't do `hi! link Delimiter DraculaPink` rather than `hi! link sassInterpolationDelimiter DraculaPink`?

The reason for that is because not every syntax uses the term `Delimiter` correctly. So this would lead to discrepancies. 

This answer applies also to your comment below.

benknoble: Actually I meant the opposite: why not `hi! link sassInterpolationDelimiter Delimiter` ?

dsifford: See 4596686 for a small last-minute adjustment that I made that changes the new sass tweaks so that instead of pointing to `Draculaxxx`, it instead links to the css equivalent, if available.

Gonna go ahead and merge since I didn't add anything beyond that.

PR #75, opened 2018-04-22
add support for fzf

dsifford:
This adds support for [fzf](https://github.com/junegunn/fzf) (see also: [fzf-vim](https://github.com/junegunn/fzf.vim/)) if the user has it installed.

The implementation only applies if the user has fzf installed and enabled. To get this to work properly, the implementation needed to occur in `after/plugin`, hence the new runtime folder additions.

**Before:**
image

**After:**
image

cc: @benknoble

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -1,0 +1,26 @@
+" Fzf: {{{
+if exists('g:loaded_fzf') && ! exists('g:fzf_colors')
+  let g:fzf_colors = {
+        \ 'fg':      ['fg', 'Normal'],
+        \ 'bg':      ['bg', 'Normal'],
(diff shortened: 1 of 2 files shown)


benknoble:
Well, I don't use `fzf`, but the screenshots look beautiful. 👍 

On a slight cosmetic note, why the movement of `netrw` code from one spot to another?

dsifford:
> On a slight cosmetic note, why the movement of netrw code from one spot to another?

Because I'm super anal about alphabetical order. Not sure how I missed that initially, but after seeing it my brain began hurting and my eye started twitching, so I had to fix it! :stuck_out_tongue_closed_eyes: 

Good to merge?

PR #76, opened 2018-04-25
Always set background colour when running gui.

siwatts:
Setting 'g:dracula_colorterm' to '0' to get a transparent background in the terminal also sets the gui background to 'none' and breaks gvim. (The gvim background ends up white).

Changed the call to the Normal highlight group to always set the background colour if the gui is running, regardless of the terminal setting.

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -225,7 +225,7 @@ call s:h('DraculaDiffDelete', s:red, s:bgdark)
 
 " Core: {{{2
 set background=dark
-call s:h('Normal', s:fg, g:dracula_colorterm == 1 ? s:bg : s:none)
+call s:h('Normal', s:fg, ( (g:dracula_colorterm == 1) || has("gui_running") ) ? s:bg : s:none)
 
 hi! link Visual DraculaSelection
 hi! link VisualNOS Visual
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -156,6 +156,14 @@ function! s:h(scope, fg, ...) " bg, attr_list, special
   execute join(l:hl_string, ' ')
 endfunction
 
+function! s:Background()
+  if g:dracula_colorterm || has("gui_running")
+    return s:bg
+  else
+    return s:none
+  endif
+endfunction
+
 "}}}2
 " Dracula Highlight Groups: {{{2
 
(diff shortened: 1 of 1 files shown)


dsifford:
Thanks for the contribution @siwatts 

LGTM .... IIRC, if running neovim `has('gui_running')` will always be 0, but I don't think this should matter.

I'll let @benknoble pull the trigger on this once he has a chance to peek at it.

siwatts:
@benknoble something like this? I put the function in the Script Helpers section.

dsifford:
I'm fine with the changes now, but I'd generally argue against abstraction until the time comes that you need to use the abstraction in more than one location. Rationale being that either way you're going to be reading the code required in the abstraction, but in cases where it's only used in one location, there's _more_ work to scroll up or move to wherever the abstraction is living, rather than just reading it in place.

No worries though. Your changes look good to me.

I'll wait on @benknoble for the final go-ahead.

siwatts:
Also @dsifford I just looked into neovim. Turns out neovim doesn't set `has("gui_running")` since it's separated into server and client and guis can attach and detach at will.

I did look into another fix, setting `s:bg` to 
```
let s:bg        = ['#282A36', 'NONE']
```
if transparency is desired keeps the hex colour for `guibg` and sets `ctermbg` to `none` for terminals, but unfortunately it breaks terminal vim in truecolor mode. I suspected truecolor vim would look at `guibg` instead of `ctermbg` and not be transparent, but for some reason the background goes the same grey as 256-color dracula. (Unpatched and this pull request both work fine in truecolor mode of course).

I don't know enough about how nvim does colourschemes or its truecolor mode to know if that would even fix nvim-qt anyway. I may install the gui for neovim and give it a play. If the same bug exists there it may be worth opening an issue on that.

dsifford:
>  I may install the gui for neovim

This doesn't exist. Or, at least, none that are "official". I'm pretty sure that neovim will always return 0 for `has('gui_running')` regardless if the user is using an external gui application.

The logic flow of how the background is currently set produces 3 potential values.

1. `g:dracula_colorterm = 1` (the default) **AND** `termguicolors` **OR** `has('gui_running')` => `#282A36`
2. `g:dracula_colorterm = 1` (the default) **AND** **NOT** (`termguicolors` **OR** `has('gui_running')`) => `236` (the closest 256 color available)
3. `g:dracula_colorterm = 0` => `NONE`

benknoble:
@dsifford Agreed about unnecessary abstraction.

My point here was more one of readability and maintainability. That three case diagram might actually be useful to have as documentation, and it wasn't 100% evident from reading the code (the logic for it is *just* complicated enough to make me scratch my head).

At any rate, has my approval.

benknoble:
Going to go ahead and merge, since we haven't found any issues and have approval from @dsifford as well.

PR #86, opened 2018-05-24
Fix end tag in xml not being highlighted

raulchen:
This is because, by default, vim links xmlTag to Function, xmlEndTag to Identifier.

before
screen shot 2018-05-24 at 7 20 07 pm
after
screen shot 2018-05-24 at 7 20 49 pm

### DIFF
--- a/after/syntax/xml.vim
+++ b/after/syntax/xml.vim
@@ -1,0 +1,7 @@
+if dracula#should_abort('xml')
+    finish
+endif
+
+hi! link xmlTag DraculaGreen
+hi! link xmlEndTag DraculaGreen
+
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -326,8 +326,6 @@ hi! link helpHyperTextJump DraculaLink
 hi! link helpCommand DraculaPurple
 hi! link helpExample DraculaGreen
 
-hi! link xmlEndTag xmlTag
-
 "}}}
 
 " vim: fdm=marker ts=2 sts=2 sw=2:
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -326,6 +326,8 @@ hi! link helpHyperTextJump DraculaLink
 hi! link helpCommand DraculaPurple
 hi! link helpExample DraculaGreen
 
+hi! link xmlEndTag xmlTag
+
 "}}}
 
 " vim: fdm=marker ts=2 sts=2 sw=2:


dsifford:
Good eye, @raulchen!

What are your thoughts on linking both `xmlTag` and `xmlEndTag` to `Keyword` instead so that it's more in line with HTML? And expanding on that, are there any other highlight groups that could be modified in XML to make the XML and HTML syntaxes line up more closely?

Would you be interested in championing this?

raulchen:
Hi @dsifford,

Thanks for your feedback. Making xml align better with html is definitely a good idea. 
But there's a problem, currently, html only highlights keywords, not tags, which means that the brackets(<>), slash (/), and custom tags are not highlighted (See the screenshot). 
Thus, if we also link xmlTag/xmlEndTag to Keyword, the entire tag will be pink. The look is still different with html.
screen shot 2018-05-25 at 11 26 06 am
Also, in terms of other groups, I am not a heavy user of xml/html, and didn't notice anything strange so far.

raulchen:
Hi @dsifford, since aligning xml with html needs more consideration, do you think it's okay to merge this fix first?

benknoble:
## html

So technically, we can "fix" the html brackets syntax with

```vim
" htmlTagName is the span in <span> *and* </span>
" currently linked to htmlStatement -> Statement -> DraculaPink
hi link htmlTag htmlTagName " htmlTag is the <> on an open tag
hi link htmlEndTag htmlTagName " htmlEndTag is the </> on a close tag
```

The original syntax for htmlTag was `Function`, but the explicit choice was made to highlight as `DraculaFg`. The original syntax for htmlEndTag is `Identifier` which points to `DraculaFg` as well.

---

## xml

I definitely think it's an issue for another discussion though, and would rather focus on xml tags here.

`xmlTag` and `xmlTagName` are both currently linked to `Function -> DraculaGreen`. These have the same purposes in html (former is the `<>`, latter is `foo` in `<foo>`).

`xmlEndTag` however is different, linked to `Identifier -> DraculaFg`. It covers the entire `</foo>` syntax (i.e. brackets and tag name).

### Suggestions

1. Merge this as a temporary fix to highlight `xmlEndTag`
2. File a bug with the xml syntax maintainer to get xmlTagName to capture the tag name in end tags as well (similar to html). Perhaps we should refer them to this post, since I provide some details on the html tag syntax groups.
3. Once that is fixed, update the highlights to be more in line with html as follows
```vim
hi link xmlTag DraculaFg
hi link xmlEndTag DraculaFg
hi link xmlTagName Keyword
```

@dsifford @raulchen let me know what you think; I can merge per (1) and open the issue per (2) above if we think this is a good plan.

dsifford:
@benknoble Sounds good to me. Care to lead the charge on this?

Also :+1: on merging this for now.

benknoble:
@dsifford will do

PR #87, opened 2018-05-25
Added extra diff groups for adds/deletes.

fourwood:
I'm not sure why this is the case, but even though the Vim help pages claim that "DiffAdd" and "DiffDelete" are the groups for diff coloring, the "DiffAdd" and "DiffDelete" links in Dracula don't seem to do anything.  Changing them to "DiffAdded" and "DiffRemoved", respectively, seem to produce the desired behavior.  I've just added an extra layer of links (e.g. linking "DiffAdded" to the pre-existing "DiffAdd") rather than changing what's there already just in case removing the originals were to break something somewhere, if that seems like a reasonable way to go about it.

(The other two diff groups, DiffChange and DiffText, seem to be right.)

Before:
diff_pre

After:
diff_post

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -243,8 +243,10 @@ hi! link ColorColumn  DraculaSelection
 hi! link CursorColumn DraculaSelection
 hi! link CursorLineNr DraculaYellow
 hi! link DiffAdd      DraculaGreen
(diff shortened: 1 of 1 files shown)


dsifford:
Hah, very strange indeed. LGTM.

Thanks for the contribution!

PR #92, opened 2018-06-20
improve airline#tabline highlighting

dsifford:
This PR is fairly simple. The additions are just a few small adjustments to the tabline highlighting for airline.

The default tabline highlighting, in my opinion, was both atrocious looking and also somewhat confusing (specifically, it was difficult to easily see what tab was the active tab)

Have a look at this and let me know what you think @benknoble.

One thing to note: I left out `airline_tabtype`, `airline_tabhid`, `airline_tablabel`, and `airline_tabmod_unsel` groups because I couldn't figure out what it was they were for. I played around with them and wasn't able to see any difference. Let me know if there's something I'm missing here and I'll fix those as well.

### DIFF
--- a/autoload/airline/themes/dracula.vim
+++ b/autoload/airline/themes/dracula.vim
@@ -107,6 +107,21 @@ let g:airline#themes#dracula#palette = {
 \}
 
 " Extensions: {{{
+" Tabline: {{{
+if get(g:, 'airline#extensions#tabline#enabled', 0)
+  let g:airline#themes#dracula#palette.tabline = {
+        \ 'airline_tabfill': s:clr('bg', 'bg'),
+        \
+        \ 'airline_tab': s:clr('comment', 'bg'),
+        \ 'airline_tabsel': s:clr('white', 'bg', 'bold'),
+        \ 'airline_tabmod': s:clr('green', 'bg'),
+        \
+        \ 'airline_tab_right': s:clr('comment', 'bg'),
+        \ 'airline_tabsel_right': s:clr('white', 'bg', 'bold'),
+        \ 'airline_tabmod_right': s:clr('green', 'bg'),
+        \}
+endif
+"}}}
 " CtrlP: {{{2
 if exists('g:loaded_ctrlp')
   let g:airline#themes#dracula#palette.ctrlp = airline#extensions#ctrlp#generate_color_map(


dsifford:
### Before

image


### After

image

dsifford:
Relevant code: https://github.com/vim-airline/vim-airline/blob/08570b6dcb496f09be842c1746ff818224bcc9df/autoload/airline/extensions/tabline.vim#L68-L112

aalvarado:
There's no option to turn the boldness on the active buffer on tabline, even if you set the dracula options to no bold. I think it's inconsistent with people who use pixel fonts, since it makes them blurry you can check #97 for how that looks.

As for colors, a middle ground for me would be a slight highlight. can I suggest using the background color of the inactive buffers from the first image as the highlight background color of the current active buffer and leave everything else the same?

I can try and contribute something but I'm very unfamiliar with Vim color schemes

dsifford:
@aalvarado Good suggestions/points made.

Just to clarify: You have bold successfully disabled everywhere else in the theme currently and bold is still persisting in the tabline? If so, yeaaaaaap, we'll need to fix that.

If you're not wanting to commit changes to the scheme directly due to unfamiliarity with vim colorschemes, perhaps maybe could you just MS Paint a few mockup ideas and paste them here for us to mull over?

aalvarado:
sure, let me check it out!!

aalvarado:
I noticed this comment from @benknoble 

> One thing I noticed (that I like) is that this looks rather similar to the default tabline colors with dracula/vim.

If you disable the tabline with `let g:airline#extensions#tabline#enabled = 0`

it looks like this:

screen shot 2018-07-23 at 10 29 59 am

seems very similar to what it got changed to, the only difference being the `tabs >` indicator and a separator between the inactive tabs, but not current, seems more confusing...

screen shot 2018-07-23 at 10 44 28 am

I think people have certain expectations on what the airline tabline should look like and seems that the style you chose could be achieved just by disabling the tabline. 

I'm thinking of something like this:

```
if get(g:, 'airline#extensions#tabline#enabled', 0)
  let g:airline#themes#dracula#palette.tabline = {
        \ 'airline_tabfill': s:clr('bg', 'bglight'),
        \
        \ 'airline_tab': s:clr('comment', 'bg'),
        \ 'airline_tabsel': s:clr('bg', 'purple'),
        \ 'airline_tabmod': s:clr('green', 'bg'),
        \
        \ 'airline_tab_right': s:clr('comment', 'bg'),
        \ 'airline_tabsel_right': s:clr('fg', 'bg', ),
        \ 'airline_tabmod_right': s:clr('green', 'bg'),
        \}
endif
```

screen shot 2018-07-23 at 10 59 03 am

what do you think?

it disables the fg colors on tabs and brings back the purple to make it look like the status line

dsifford:
Good points. At the very least, the changes you suggest will serve as a decent starting point for further tweaks down the road if needed.

Feel free to send a PR for it if you're feeling up to it. Otherwise, I'll try to get to this soon(ish)

aalvarado:
will try to open up a pr today :) thanks for your help!

PR #102, opened 2018-09-09
Change specialKey from red to subtle

blmayer:
I use listings in vim, so it prints invisible characters. However the tab and the space characters are drawn in red, taking too much attention, so I decided to make them subtle. To activate the listing add this to your .vimrc file:

set listchars=eol:¬,tab:>-,trail:~,extends:>,precedes:<,space:·
set list
So you can see the red is really disguising.

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -321,7 +321,7 @@ hi! link Type DraculaCyanItalic
 hi! link Delimiter DraculaFg
 
 hi! link Special DraculaPink
-hi! link SpecialKey DraculaRed
+hi! link SpecialKey DraculaSubtle
 hi! link SpecialComment DraculaCyanItalic
 hi! link Tag DraculaCyan
 hi! link helpHyperTextJump DraculaLink


dsifford:
Thanks for the contribution @blmayer!

This had me puzzled for a few minutes because I could not reproduce this whatsoever on my end, but I think I got it sorted out. 

Apparently, the `Whitespace` group (which we have set subtle) is a highlight group extension that is _only_ defined in neovim. So yeah, I'm surprised that it took this long for someone to complain about this !:stuck_out_tongue_closed_eyes:

Your fix works just fine for now but I'll want to eventually go back soon and adjust so that neovim users still have `SpecialKey` set at `DraculaRed` (which is helpful when sending control codes / escape sequences).

blmayer:
Thank you for accepting my PR! 

> So yeah, I'm surprised that it took this long for someone to complain about this !stuck_out_tongue_closed_eyes

Me too, I had to fix it as soon as I saw the red, perhaps few people uses vim with listing.

benknoble:
I use vim with `'list'` day-to-day, but didn't frankly mind the red. Unfortunately, it seems vim uses `SpecialKey` for those characters...

dsifford:
@benknoble Assuming you use neovim? Or are you saying you prefer the red characters for whitespace invisibles?

benknoble:
Regular vim, actually. Simply that I was used to the red and it never bothered me. 'Prefer' is a strong word :P

PR #103, opened 2018-09-12
Remove italics on Ruby constants

pachonk:
Before:
image

After:
image

### DIFF
--- a/after/syntax/ruby.vim
+++ b/after/syntax/ruby.vim
@@ -14,3 +14,4 @@ hi! link rubyInstanceVariable       DraculaPurpleItalic
 hi! link rubyInterpolationDelimiter DraculaPink
 hi! link rubyRegexpDelimiter        DraculaRed
 hi! link rubyStringDelimiter        DraculaYellow
+hi! link rubyConstant               DraculaPurple


pachonk:
https://github.com/pachonk/.dotfiles/commit/c49612abc16e11c9aee8442cf8ac12c32d18c76f

dsifford:
Yep! I fought with that as well (specifically, and IIRC, exclusively, on Mac OS using iTerm)... 

Glad you got it sorted out.

Side note: If you happen to be using `^2.1` and are not running Darwin + iTerm2, you can actually just set that to `tmux-256color` and it should work out of the box.

dsifford:
Correct. That's what I have [set for mine](https://github.com/dsifford/.dotfiles/blob/master/config/tmux/tmux.conf#L8).

Also, you may need to explicitly [force truecolor support](https://github.com/dsifford/.dotfiles/blob/master/config/tmux/tmux.conf#L15) to get rich colors.

And finally, the following is a handy function you might want to use to ensure your terminal has all the goodies supported: https://github.com/dsifford/.dotfiles/blob/master/local/lib/bashrc.d/functions/terminal-test.sh

PR #104, opened 2018-09-19
adjust xml so that it more closely matches html

dsifford:
This PR adjusts xml so that it more closely matches what HTML looks like.

Pretty simple stuff, but figured I should get a review on it.

(before and after pics to follow)


benknoble:
Just tested in `vim`, no problems on this end. Feel free to merge.

Issue #91, opened 2018-06-17
Error is not distinguishable.


zhou13:
I found the dracula's error color is not very distinguishable:
![image](https://user-images.githubusercontent.com/576382/41512983-1bd37460-7249-11e8-8c17-b4be9ec68709.png)

In fact, `b // asdf` does not look like an error but a keyword!

On the other hand, default theme and desert clearly marked it:
![image](https://user-images.githubusercontent.com/576382/41513000-711a7d24-7249-11e8-8865-a095165325a1.png)
![image](https://user-images.githubusercontent.com/576382/41513003-7d45704a-7249-11e8-8055-8ffb55f7dc33.png)

Currently I put the workaround `hi! DraculaError gui=inverse` or `hi! DraculaError gui=undercurl`
in my vimrc and I get
![image](https://user-images.githubusercontent.com/576382/41513068-b1394902-724a-11e8-86e6-c2dc1433712e.png)
![image](https://user-images.githubusercontent.com/576382/41513085-0f53a5c8-724b-11e8-8712-2d212cc27098.png)

dsifford:
Are these errors being produced by the syntax, or are you using a language server or linting library?

If the latter, the library should provide the undercurls (see #84)

Regarding the color: That's the dracula-defined error color. We can't stray from it. But as you've alluded, you're more than welcome to modify it in your init scripts.

Re: inverting the color -- Same as above IMHO. I don't think most would want the errors to be that extremely colored, especially those that use aggressive, real-time linters or language servers.

Thoughts @benknoble?

zhou13:
Those error are produce by syntax.  Also I have the same problem with `spell`.  Not sure how can I add underline without touching the highlight theme.

dsifford:
cc @nudzg

benknoble:
@zhou13 `spell` uses `SpellBad` which is linked to `DraculaErrorLine`

zhou13:
I found the reason that the library should provide the undercurls does not make a lot of sense to me because vim syntax marked error as Error. Even for ALE the default setting in dracula is not nice due to the usage of TODO.

Here is how I patch dracula to make both ALE and syntax error look nicer:
```
hi! link Error DraculaErrorLine

hi! link ALEError DraculaErrorLine
hi! link ALEWarning DraculaWarnLine
hi! link ALEInfo DraculaInfoLine

hi! link ALEErrorSign DraculaRed
hi! link ALEWarningSign DraculaOrange
[...]

Issue #84, opened 2018-05-18
Undercurl should not be forced by color scheme in errors/warnings


nudzg:
### What happened

Using w0rp/ale to display errors/warnings I see the indicator on left gutter has an unexpected undercurl. 

### What I expected to happen

Left gutter indicator should not be undercurled.

Setting undercurl seems like is something ale (or equivalent) has to do, not the colorscheme.

If I change:

call s:h('DraculaError', s:red, s:none, [s:attrs.undercurl], s:red)

to:

call s:h('DraculaError', s:red, s:none, [], s:red)

I get what I think is correct.

### Screenshot
![dracula-ale](https://user-images.githubusercontent.com/5119900/40226833-18430668-5a8d-11e8-9cb4-062eeee70f80.png)

Top example is the current and in my opinion wrong behavior. Bottom is what I get with the suggested change.

### Machine Info
- **Vim type (`vim`/`gvim`/`neovim`)**: gvim
- **Vim version**: 8.0.1815
- **OS**: Linux xxx 4.16.8-1-ARCH #1 SMP PREEMPT Wed May 9 11:25:02 UTC 2018 x86_64 GNU/Linux

dsifford:
We can give your suggestions a shot. Sounds reasonable.

PR #109, opened 2018-10-22
Patch ALE highlights to use dracula colors

benknoble:
By default, they are:

- ALEError => SpellBad => DraculaErrorLine
- ALEWarning => SpellCap => DraculaInfoLine
- ALEInfo => ALEWarning => SpellCap => DraculaInfoLine

- ALEErrorSign => Error => DraculaError
- ALEWarningSign => Todo => DraculaTodo
- ALEInfoSign => ALEWarningSign => Todo => DraculaTodo

Proposed change:

- ALEError => DraculaErrorLine
- ALEWarning => DraculaWarningLine
- ALEInfo => DraculaInfoLine

- ALEErrorSign => DraculaError
- ALEWarningSign => DraculaWarning
- ALEInfoSign => DraculaInfo

From discussion in #91

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -36,7 +36,7 @@ endif
 " ALE: {{{
 if exists('g:ale_enabled')
   hi! link ALEError DraculaErrorLine
-  hi! link ALEWarning DraculaWarningLine
+  hi! link ALEWarning DraculaWarnLine
   hi! link ALEInfo DraculaInfoLine
 
   hi! link ALEErrorSign DraculaError
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -33,4 +33,16 @@ endif
 
 "}}}
 
+" ALE: {{{
+if exists('g:ale_enabled')
+  hi! link ALEError DraculaErrorLine
+  hi! link ALEWarning DraculaWarningLine
+  hi! link ALEInfo DraculaInfoLine
+
+  hi! link ALEErrorSign DraculaError
+  hi! link ALEWarningSign DraculaWarning
+  hi! link ALEInfoSign DraculaInfo
+endif
+" }}}
+
 " vim: fdm=marker ts=2 sts=2 sw=2:
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -39,9 +39,9 @@ if exists('g:ale_enabled')
   hi! link ALEWarning DraculaWarnLine
   hi! link ALEInfo DraculaInfoLine
 
-  hi! link ALEErrorSign DraculaError
-  hi! link ALEWarningSign DraculaWarning
-  hi! link ALEInfoSign DraculaInfo
+  hi! link ALEErrorSign DraculaRed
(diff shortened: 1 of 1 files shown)


benknoble:
(untested)

after/plugin/dracula.vim
```diff
@@ -33,4 +33,16 @@ endif
 
 "}}}
 
+" ALE: {{{
+if exists('g:ale_enabled')
+  hi! link ALEError DraculaErrorLine
+  hi! link ALEWarning DraculaWarningLine
```
dsifford:
```suggestion
  hi! link ALEWarning DraculaWarnLine
```

after/plugin/dracula.vim
```diff
@@ -33,4 +33,16 @@ endif
...
+  hi! link ALEInfo DraculaInfoLine
+
+  hi! link ALEErrorSign DraculaError
```
dsifford:
```suggestion
```

dsifford: This is already linked to `Error` which is better than setting explicitly in the event that the person changes themes.

after/plugin/dracula.vim
```diff
@@ -33,4 +33,16 @@ endif
...
+  hi! link ALEWarningSign DraculaWarning
```
dsifford:
```suggestion
```

dsifford: `DraculaWarning` doesn't exist currently.

after/plugin/dracula.vim
```diff
@@ -33,4 +33,16 @@ endif
...
+  hi! link ALEInfoSign DraculaInfo
```
dsifford:
```suggestion
```

dsifford: `DraculaInfo` doesn't exist currently.

dsifford:
In all honesty, I'd prefer to just leave as-is, since explicitly setting error colors will cause issues when users switch themes. But your call. No strong feelings either way.

dsifford:
Also check out #84 for potentially relevant issues.

benknoble:
I'm not sure which "potential issues" you mean by #84 , but I do see your point that setting `ALE` groups shouldn't be our responsibility (and breaks changing colorschemes, but I think we've discussed that before? I don't remember our end conclusions about that).

It's just unfortunate that our color scheme seems to be 'off' in the sense that warnings look like info.

dsifford:
I was referring to the underlines (or in this case undercurls)... Not sure if these changes would reopen that issue or not.

I'm indifferent though. If you checked it and this fixes the warn/info highlights on your end, then it's cool by me. We can always fix again later if something else comes up. 👍

dsifford:
Up to you. I don't have strong feelings either way. If you dig the changes merge it in and if something breaks, we'll fix it.

benknoble:
Alright. I'll merge it in and do some light testing as I work the next week or so; if something comes up we can easily revert the merge commit.

dsifford:
Sounds good! I'm a devout ALE user myself, so if you don't catch something, I will :relaxed:

PR #113, opened 2018-11-14
Small misc tweaks/fixes

dsifford:
This PR consists of 3 things:

1. Improvements of `MatchParen` highlighting. (see images below)
2. Small tweaks to `CursorColumn` and `ColorColumn` (make more subtle & less in the way).
3. Make `CursorLine` optional.

TODO: screenshots.

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -119,6 +119,10 @@ if !exists('g:dracula_colorterm')
   let g:dracula_colorterm = 1
 endif
 
+if !exists('g:dracula_cursorline')
+  let g:dracula_cursorline = 1
+endif
+
 "}}}2
 " Script Helpers: {{{2
 
@@ -237,7 +241,12 @@ call s:h('Normal', s:fg, s:Background())
 call s:h('StatusLine', s:none, s:bglighter, [s:attrs.bold])
 call s:h('StatusLineNC', s:none, s:bglight)
 call s:h('WildMenu', s:bg, s:purple, [s:attrs.bold])
-call s:h('CursorLine', s:none, s:subtle)
+
+if g:dracula_cursorline
+  call s:h('CursorLine', s:none, s:subtle)
+else
+  hi clear CursorLine
+endif
 
 hi! link ColorColumn  DraculaBgDark
 hi! link CursorColumn DraculaBgDark
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -239,8 +239,8 @@ call s:h('StatusLineNC', s:none, s:bglight)
 call s:h('WildMenu', s:bg, s:purple, [s:attrs.bold])
 call s:h('CursorLine', s:none, s:subtle)
 
-hi! link ColorColumn  DraculaSelection
-hi! link CursorColumn DraculaSelection
+hi! link ColorColumn  DraculaBgDark
+hi! link CursorColumn DraculaBgDark
 hi! link CursorLineNr DraculaYellow
 hi! link DiffAdd      DraculaGreen
 hi! link DiffAdded    DiffAdd
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -276,7 +276,7 @@ hi! link WarningMsg   DraculaOrangeInverse
 " Syntax: {{{
 
 " Required as some plugins will overwrite
-call s:h('MatchParen', s:fg, s:pink, [s:attrs.underline])
+call s:h('MatchParen', s:green, s:none, [s:attrs.underline])
 call s:h('Conceal', s:comment, s:bglight)
 
 " Neovim uses SpecialKey for escape characters only. Vim uses it for that, plus whitespace.
--- a/doc/dracula.txt
+++ b/doc/dracula.txt
@@ -56,25 +56,29 @@ In the following section, `1` signifies `on` and `0` signifies `off`.
 * *g:dracula_bold*
 Include bold attributes in highlighting >
     let g:dracula_bold = 1
-<
+
 * *g:dracula_italic*
 Include italic attributes in highlighting >
     let g:dracula_italic = 1
-<
+
 * *g:dracula_underline*
 Include underline attributes in highlighting >
     let g:dracula_underline = 1
-<
+
 * *g:dracula_undercurl*
 Include undercurl attributes in highlighting (only if underline enabled) >
     let g:dracula_undercurl = 1
-<
+
 * *g:dracula_inverse*
 Include inverse attributes in highlighting >
     let g:dracula_inverse = 1
 
 * *g:dracula_colorterm*
 Include background fill colors >
+   let g:dracula_colorterm = 1
+
+* *g:dracula_cursorline*
+Include cursor line highlighting (otherwise just highlight line number) >
    let g:dracula_colorterm = 1
 
 ==============================================================================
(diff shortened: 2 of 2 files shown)


colors/dracula.vim
```diff
@@ -237,10 +241,15 @@ call s:h('Normal', s:fg, s:Background())
...
-hi! link ColorColumn  DraculaSelection
-hi! link CursorColumn DraculaSelection
+if g:dracula_cursorline
+  call s:h('CursorLine', s:none, s:subtle)
+else
+  hi clear CursorLine
+endif
+
+hi! link ColorColumn  DraculaBgDark
+hi! link CursorColumn DraculaBgDark
```
benknoble: `ColorColumn` is indeed subtle (👍 ); I think `CursorColumn` is a bit _too_ subtle (I had to search hard to find it).

colors/dracula.vim
```diff
@@ -237,10 +241,15 @@ call s:h('Normal', s:fg, s:Background())
 call s:h('StatusLine', s:none, s:bglighter, [s:attrs.bold])
 call s:h('StatusLineNC', s:none, s:bglight)
 call s:h('WildMenu', s:bg, s:purple, [s:attrs.bold])
-call s:h('CursorLine', s:none, s:subtle)
 
-hi! link ColorColumn  DraculaSelection
-hi! link CursorColumn DraculaSelection
+if g:dracula_cursorline
```
benknoble: I'm not sure I follow why we conditionally define `CursorLine`; I thought the idea was to define it to Dracula's colors and let user's decide whether or not they turn it on. I've recently started working without it entirely, but this change doesn't make any sense to me.

Effectively, this feels weird and non-intuitive:

1. `let g:dracula_cursorline = 0` in vimrc
2. Start vim
3. Turn on cursorline, expecting to get a (probably non-Dracula) cursorline... and there isn't one? It's just... gone?

Without highlighting, it almost doesn't matter whether I have it on or off (and it seems like we're implicitly making that decision for the user).

I'd rather just leave the highlighting the way it is, and let the user clear it if they want, turn off/on if they want, etc. 

Let me know if my thoughts here are confusing; I'm finding it hard to express quite why I don't like this. It just seems like setting that variable disables cursorline entirely.

colors/dracula.vim
```diff
@@ -276,7 +285,7 @@ hi! link WarningMsg   DraculaOrangeInverse
 " Syntax: {{{
 
 " Required as some plugins will overwrite
-call s:h('MatchParen', s:fg, s:pink, [s:attrs.underline])
+call s:h('MatchParen', s:green, s:none, [s:attrs.underline])
```
benknoble: I like this new version in the screenshot, mostly because it makes it more clear which paren the cursor is on. I could go either way for the colors (green v. pink).

doc/dracula.txt
```diff
@@ -56,19 +56,19 @@ In the following section, `1` signifies `on` and `0` signifies `off`.
 * *g:dracula_bold*
 Include bold attributes in highlighting >
     let g:dracula_bold = 1
-<
+
```
benknoble: Not sure these are necessary but /shrug

colors/dracula.vim
dsifford: Nope, your thoughts are valid.

The reason for the conditional definition here is **exclusively** due to the fact that some people (me) want a situation where the cursor line is not highlighted, **BUT** the line number of the cursor line is highlighted in all modes. That only works by doing it this way. (see `:h hl-CursorLineNr` for more info)

colors/dracula.vim
dsifford: Suggestions?

colors/dracula.vim
dsifford: I used the green for about a week and it grew on me. The pink didn't work as well for me because it wasn't _quite_ as vibrant enough. That's why I switched it.

doc/dracula.txt
dsifford: `<` is only necessary if you want to end the code block on a line that doesn't begin on column 1. Removed just because they were confusing when i went in and added the new option.

colors/dracula.vim
benknoble: Ahh. I see... hm. Can that be written somewhere besides this issue to tie it in better as justification? Have we thought about different solutions? Could we perhaps add a note to the documentation for users that might want e.g.:

```vim
autocmd ColorScheme dracula hi clear CursorLine
```

to solve the problem? This seems like a non-dracula problem to me regardless.

I use relative number so I wasn't aware of that problem (the highlighting is always active with relativenumber on).

doc/dracula.txt
benknoble: Got it.

colors/dracula.vim
dsifford: ^^^I like that idea muuuuuuuch better. I wasn't aware `ColorScheme` was an event. Good deal. I'll remove this stuff

colors/dracula.vim
benknoble: I'm not sure. The default red seems ugly. I just don't use it to know what to look for.

dsifford: It's showing up red for you?

dsifford: My vim -u NONE is the same color as my terminal foreground color.

I don't use it either, but I can't imagine something as noticable as the foreground color would be very useful in any situation. The dark on my terminal is not difficult to see at all, which makes me wonder why it is on yours?

benknoble: Sorry I got mixed up between cursor column and color column. Ignore the red.

The dark on my terminal can be difficult to see because I use our Terminal.app theme as well, so the background is already quite dark.

dsifford:
@benknoble We good? Thoughts?

benknoble:
Yes, I think we are okay to merge. I don't really use the column/line features that were darkened; if we get backlash, I think we can consider reverting it.

bdielman:
Agree with @benknoble  that the change for `ColorColumn` is good. Subtle, but still noticeable. I preferred the `CursorColumn` before as it's way too subtle now. I feel the `CursorColumn` and `CursorLine` colors should be similar as they are in other themes, like Solarized.

The `MatchParen` change would work better for me if the both parens would look like they do in @dsifford's example, but the background color of the opposing paren has a white background instead of green. Not sure if it's something with my settings. Currently using OS X terminal with the Dracula theme.

dsifford:
@bdielman Happy to revert and/or adjust `CursorColumn`. I don't really use it, but in the one instance that I did, it felt like the plain white was too much.

Can you post a screenshot of your `MatchParen`? Not sure I understand what you're saying.

benknoble:
@bdielman the green-background-ed paren in the screenshot is the one with the cursor on it

bdielman:
@dsifford @benknoble - Here's a screenshot. Not a huge deal, just curious how to get the bg color to match the green in the screenshot above. Again, could be something with my setup.

matchparens

dsifford:
@bdielman ah, looks like `CursorColumn` is overwriting the foreground color. That should also be fixed. Not sure if that’s gonna be totally fixable though.

bdielman:
@dsifford - Thanks!

bdielman:
Removed `set cursorcolumn` from my vimrc and the foreground color is still white. Again, not a big deal so if it's not an easy fix I can live with it.

benknoble:
@dsifford  I can't find anywhere that `DraculaBgDark` (which `CursorColumn` links to) sets the foreground, but I didn't look very far.

dsifford:
Figured it out... You were right @benknoble, it was not an overwrite of the cursor foreground color, but instead a quirk on Mac OS.

I'm not sure how to fix it for Terminal.app @bdielman, however, I do know how to fix it on iTerm2.

**iTerm2 tl;dr:** `Preferences->Profiles->Colors` and then check `Smart box cursor color`.

### Before

image

### After

image

bdielman:
@dsifford - Awesome! Thanks for looking into this!

benknoble:
Not investigated whether it occurs on Terminal.app yet

dsifford:
@benknoble it does

PR #120, opened 2018-12-27
Adapt identifier and airline style

saschagrunert:
I would like to propose this design change to have a better look and feel in my modest opinion.

You can see in the lower left side of the screenshot that the following changed:

- identifiers are now orange and provide a better visual feedback
- the StatusLine fits into the background and seems less noisy

What do you think about it?

### DIFF
--- a/autoload/airline/themes/dracula.vim
+++ b/autoload/airline/themes/dracula.vim
@@ -46,7 +46,7 @@ let g:airline#themes#dracula#palette = {
 \   'normal': s:color_map(
 \       ['bg', 'purple'],
 \       ['fg', 'comment'],
-\       ['fg', 'selection'],
+\       ['fg', 'bg'],
 \       {
 \         'airline_warning': s:clr('bg', 'orange'),
 \         'airline_error': s:clr('bg', 'red'),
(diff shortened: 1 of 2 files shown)


dsifford:
Thanks for the contribution @saschagrunert.

I do understand and get that the changes add a bit more contrast to Go syntax, but adding this change would move the implementation [out of spec alignment](https://dsifford.github.io/dracula-spec/#sec-Variables) with the other implementations.

The issue with Vim specifically is that its current method of applying syntax highlighting is far more rudimentary to most other editors, so its difficult and/or impossible to scope syntax regions appropriately (see `len`, `append`, and `make` in your example above).

Neovim is currently looking into possibly adding [tree-sitter syntaxes](https://github.com/neovim/neovim/issues/1767#issuecomment-353728712) and dropping the legacy syntax parser which would completely resolve this issue. Not sure where vim proper stands on this.

In short: If this is a must for you, just add your changes into your vim init scripts. 

Sorry for the bad news. Hope you can understand!

saschagrunert:
Alright, thank you for the detailed explanation. 🙏

PR #121, opened 2019-01-02
Add syntax files for Perl 5 and Perl 6

mienaikage:
Some tiny additions to distinguish variables.

### DIFF
--- a/after/syntax/perl.vim
+++ b/after/syntax/perl.vim
@@ -1,0 +1,5 @@
+if dracula#should_abort('perl')
+    finish
+endif
+
+hi! link perlVarPlain DraculaCyan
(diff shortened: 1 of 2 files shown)


benknoble:
These changes would unfortunately take the vim theme out of line with the [spec](https://dsifford.github.io/dracula-spec/#Variable).

As noted in our help document, you can "hot-patch" this yourself if you desire by placing similar code in e.g. `~/.vim/after/ftplugin/perl/dracula.vim`:

```vim
augroup dracula_perl_override
  au!
  au ColorScheme dracula hi! link perlVarPlain DraculaCyan
augroup END
```

(Make sure to update `b:undo_ftplugin` accordingly.)

mienaikage:
Ah, useful! I'd not encountered the spec before, maybe it'd be worth having on dracula/dracula-theme? Seems pygments and atom may be out of spec as that's where I got the idea from.

dsifford:
https://github.com/dracula/dracula-theme/issues/232

PR #122, opened 2019-01-10
Syntax highlighting for Perl 5 in line with spec

mienaikage:
Hash key color is as it appears in vscode.

The various statements cover the builtin functions of Perl 5.

`perl_string_as_statement` is an existing variable used in vim-perl, I wasn't 100% sure with which color to go with (it shows as pink but that's also for escape sequences), but red seemed the most fitting.

### DIFF
--- a/after/syntax/perl.vim
+++ b/after/syntax/perl.vim
@@ -1,0 +1,38 @@
+if dracula#should_abort('perl')
+    finish
+endif
+
+" Regex
+hi! link perlMatchStartEnd       DraculaRed
+
+" Builtin functions
+hi! link perlOperator            DraculaCyan
+hi! link perlStatementFiledesc   DraculaCyan
+hi! link perlStatementFiles      DraculaCyan
+hi! link perlStatementFlow       DraculaCyan
+hi! link perlStatementHash       DraculaCyan
+hi! link perlStatementIOfunc     DraculaCyan
+hi! link perlStatementIPC        DraculaCyan
+hi! link perlStatementList       DraculaCyan
+hi! link perlStatementMisc       DraculaCyan
+hi! link perlStatementNetwork    DraculaCyan
+hi! link perlStatementNumeric    DraculaCyan
+hi! link perlStatementProc       DraculaCyan
+hi! link perlStatementPword      DraculaCyan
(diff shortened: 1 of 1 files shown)


after/syntax/perl.vim
```diff
@@ -0,0 +1,38 @@
+if dracula#should_abort('perl')
+    finish
+endif
+
+" Regex
+hi! link perlMatchStartEnd       DraculaRed
+
+" Builtin functions
+hi! link perlOperator            DraculaCyan
+hi! link perlStatementFiledesc   DraculaCyan
```
dsifford: Is there perhaps a parent scope here that hits all of these (e.g. `perlStatement`)?

Would be better to just target that rather than each individually.

after/syntax/perl.vim
```diff
@@ -0,0 +1,38 @@
...
+hi! link perlStatementPword      DraculaCyan
+hi! link perlStatementRegexp     DraculaCyan
+hi! link perlStatementScalar     DraculaCyan
+hi! link perlStatementSocket     DraculaCyan
+hi! link perlStatementTime       DraculaCyan
+hi! link perlStatementVector     DraculaCyan
+
+" Highlighting for quoting constructs, tied to existing option in vim-perl
+if get(g:, 'perl_string_as_statement', 0)
+  hi! link perlStringStartEnd DraculaRed
```
dsifford: Is there a need for the conditional here?

benknoble: @dsifford not an answer to your question, but cf. `:help ft-perl-syntax`

after/syntax/perl.vim
mienaikage: I tried looking, but unfortunately perlStatement also covers control flow e.g `for`, `while` and phasers e.g `BEGIN`, `INIT` so I had to pick them out individually.

after/syntax/perl.vim
mienaikage: I didn't want to override it completely (appears yellow by default, same as the string itself), and when the option is enabled it changes to pink, which is the same color as escaped characters within strings. I decided that changing the color when the option is enabled would be the best option.

dsifford: Got it. Sounds good.

PR #131, opened 2019-03-20
Add basic support for CtrlP

bdielman:
I figure this would be a nice addition. It will highlight the matched words when typing a search using CtrlP.

### Before:
Screen Shot 2019-03-20 at 12 37 38 PM

### After:
Screen Shot 2019-03-20 at 12 31 33 PM

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -31,10 +31,11 @@ if exists('g:ale_enabled')
   hi! link ALEInfoSign DraculaCyan
 endif
 " }}}
-"}}}
 " CtrlP: {{{
-  hi! link CtrlPMatch Keyword
+if exists('g:loaded_ctrlp')
+  hi! link CtrlPMatch IncSearch
   hi! link CtrlPBufferHid Normal
+endif
 " }}}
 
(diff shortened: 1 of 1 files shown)


after/plugin/dracula.vim
```diff
@@ -31,5 +31,10 @@ if exists('g:ale_enabled')
   hi! link ALEInfoSign DraculaCyan
 endif
 " }}}
+"}}}
+" CtrlP: {{{
+  hi! link CtrlPMatch Keyword
```
dsifford: 2 quick things:

1. Can we add an `exists` check here to test for the existence of ctrlp before creating those highlight groups.
2. Can we switch from `Keyword` to `IncSearch` so there's better uniformity with other means of searching (FZF, builtin search, etc)

bdielman: Sounds good! I'll make those changes.

bdielman: Just pushed the requested changes.

dsifford:
Great work. Thanks!

bdielman:
Thanks! 🎉

PR #135, opened 2019-05-01
Modify dracula theme for my needs

saschagrunert:


dsifford:
Hi there.

These changes are better suited for a fork as the color changes are [off specification](https://dsifford.github.io/dracula-spec/). If you're passionate about these changes, you're welcome to bring them up in the [running discussion of the spec](https://github.com/dracula/dracula-theme/issues/232). 

Thanks

benknoble:
@saschagrunert See the help document for suggestions to overrule/modify the Dracula theme dynamically; it basically just uses `autocmd ColorScheme dracula ...`

@dsifford I've just noticed that our highlight link statements do not use the `default` argument, which makes overriding them potentially harder (see `:help hi-default`). I can submit a PR if you want the changes.

dsifford:
@benknoble that was done on purpose because tons of plugins don't follow the rules either. So leaving the default unset was the only way to be absolutely sure that the theme loads the same way for everybody.

Users can still change highlights in their own vimrc's if they do it after loading the theme.

PR #153, opened 2019-10-20
fix: move back to "typescriptreact" from "typescript.tsx" filetype

dsifford:
This change is in accordance to the now official filetype defined by vim itself.

See: https://github.com/vim/vim/commit/92852cee3fcff1dc6ce12387b234634e73267b22

### DIFF
--- a/after/syntax/typescript.vim
+++ b/after/syntax/typescript.vim
@@ -1,4 +1,4 @@
-if dracula#should_abort('typescript')
+if dracula#should_abort('typescript', 'typescriptreact')
     finish
 endif
 
--- a/after/syntax/typescript.vim
+++ b/after/syntax/typescript.vim
@@ -2,6 +2,8 @@ if dracula#should_abort('typescript')
     finish
 endif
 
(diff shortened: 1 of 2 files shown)


after/syntax/typescript.vim
```diff
@@ -2,6 +2,8 @@ if dracula#should_abort('typescript')
     finish
 endif
 
+" HerringtonDarkholme/yats.vim {{{
```
benknoble: Are these groups defined by this plugin? Just to clarify

dsifford: Yes. Figured I'd match the format we have going in the `javascript.vim` file.

benknoble: Ah, gotcha. I don't use these filetypes (or have the patches to vim yet), but it looks fine to me.

dsifford: Yeah, vim is a slow moving ship when it comes to them actually pushing releases. This merged to master back in August!

Cool, I'll merge this in here then. :+1:

PR #163, opened 2019-12-22
Set colors for syntax defined by stephpy/vim-yaml.

TheWall89:
An attempt to fix Dracula colors when using [stephpy/vim-yaml](https://github.com/stephpy/vim-yaml) syntax definitions. Result is not perfect, as the syntax file does not match a lot of things. Screenshots attached:

Dracula with default vim syntax for YAML:
default_vim

Dracula with stephpy/vim-yaml before the fix:
stephpy_before

Dracula with stephpy/vim-yaml after the fix:
stephpy_after

The default vim syntax seems superior, but stephpy/vim-yaml plugin is widely used, e.g. in [sheerun/vim-polyglot](https://github.com/sheerun/vim-polyglot). So, the fix might still be useful.

### DIFF
--- a/after/syntax/yaml.vim
+++ b/after/syntax/yaml.vim
@@ -10,7 +10,7 @@ hi! link yamlFlowIndicator   Delimiter
 hi! link yamlNodeTag         DraculaPink
 hi! link yamlPlainScalar     DraculaYellow
 
-" Links for stephpy/yaml
+" stephpy/yaml {{{
 
 " yamlConstant
 hi! link Keyword DraculaPurple
@@ -31,3 +31,5 @@ hi! link Operator Delimiter
 hi! link String DraculaYellow
 " yamlEscape
 hi! link Special DraculaPink
+
+" }}}
--- a/after/syntax/yaml.vim
+++ b/after/syntax/yaml.vim
@@ -10,3 +10,24 @@ hi! link yamlFlowIndicator   Delimiter
 hi! link yamlNodeTag         DraculaPink
 hi! link yamlPlainScalar     DraculaYellow
 
+" Links for stephpy/yaml
+
+" yamlConstant
+hi! link Keyword DraculaPurple
+" yamlIndicator
+hi! link PreCondit DraculaPink
+" yamlAnchor, yamlAlias
+hi! link Function DraculaPinkItalic
+" yamlKey
+hi! link Identifier DraculaCyan
+" yamlType
+hi! link Type DraculaPink
+
+" yamlComment
+hi! link Comment DraculaComment
+" yamlInline, yamlBlock
+hi! link Operator Delimiter
+" yamlString
+hi! link String DraculaYellow
+" yamlEscape
+hi! link Special DraculaPink


TheWall89:
Sure can do!
Just to be sure, by "fold" you mean this:

```
" HerringtonDarkholme/yats.vim {{{
...
" }}}
```
right?

Ps. Do you want me to remove the comments?

dsifford:
Re: fold -- correct!

Re: remove comments -- up to you. I'm cool to keep those if you find them helpful.

dsifford:
Thanks!

dsifford:
Also.. Re: vim-polyglot installs this one but it's not better than default -- might want to bring that up with the maintainers there. I'm sure they'd prob be glad to offload one if it's not needed!

dsifford:
@TheWall89 Actually, I'm gonna have to roll this merge back... I didn't look at this hard enough, but `stephpy/vim-yaml` redefines global highlight regions that we cannot redefine (e.g. Keyword, Identifier, etc).. 

Sorry about that.

benknoble:
All, after a glance at https://github.com/stephpy/vim-yaml/blob/master/after/syntax/yaml.vim, I think @TheWall89 only needs to color the following groups (instead of the "big" vim ones, like `Keyword`):

```
yamlAlias
yamlAnchor
yamlBlock
yamlComment
yamlConstant
yamlEscape
yamlIndicator
yamlInline
yamlKey
yamlString
yamlType
```

TheWall89:
@dsifford : no need to apologize for that! I was actually concerned about coloring those very generic keywords, and I was waiting for your review.

@benknoble : actually, that's the very first thing I tried but it doesn't work. For example, using

```
hi! link yamlKey DraculaCyan
```

The result is the following
yamlkey_nono

As you can see, the keys are not colored in cyan. Am I missing something here?

I am also going to open an issue on polyglot repo.

TheWall89:
Hey @dsifford,

searching for issues in sheerun/vim-polyglot I found one by you about this very thing with yaml :smile: 

https://github.com/sheerun/vim-polyglot/issues/297

The answer sheerun gave you sounds solid and I don't believe we should insist. :blush: Also because we can easily disable yaml syntax from polyglot with 

```
let g:polyglot_disabled = ['yaml']
```

benknoble:
@TheWall89 that’s very odd; I would have expected that to work without a hiccup. Can you confirm the syntax group applied to the keys?

TheWall89:
> @TheWall89 that’s very odd; I would have expected that to work without a hiccup. Can you confirm the syntax group applied to the keys?

In stephpy syntax file, yaml.vim, youcan find these lines at the end:

```
" Setup the highlighting links
hi link yamlKey         Identifier
```

Now, if I color `yamlKey` with

```
hi link yamlKey DraculaCyan
```

nothing happens. If I color `Identifier`, with

```
hi link Identifier DraculaCyan
```

YAML keys gets the cyan. I tried to switch between them and I can confirm the behavior. Closed and reopened a yaml file in vim to be sure everything was reloaded.

I'm not an expert in vim syntax/plugins and I'm proceeding by trial and error. I have no more ideas :disappointed:

benknoble:
@TheWall89 by *confirm the syntax group*, I meant inspect the syntax group: plugins like scriptease or my own synstax provide mappings to do this; it's basically a one-liner function call otherwise.

By *close and reopened*, do you mean you

1. Unloaded and reloaded the buffer; or
2. Closed and reopened a (vim) window; or
3. Quit and restarted vim?

The distinction is important.

The simplest thing to do is load an instance of vim with the stephy plugin loaded, and then manually do `:hi! link ...` and see what happens, if you're just experimenting manually.

TheWall89:
> By _close and reopened_, do you mean you
> 
> 1. Unloaded and reloaded the buffer; or
> 
>2. Closed and reopened a (vim) window; or
> 
>3. Quit and restarted vim?
> 
> 
> The distinction is important.

How I am working:

1. Open Konsole terminal emulator. 
2. Run `vim ~/.vim/bundle/vim/after/syntax/yaml.vim` (Dracula file for coloring yaml)
3. Change something, save and close with `:wq`
4. Run `vim ~/Downloads/test.yaml`
5. Close with `:q`
6. Repeat steps 2 to 5
 
> The simplest thing to do is load an instance of vim with the stephy plugin loaded, and then manually do `:hi! link ...` and see what happens, if you're just experimenting manually.

I tried this and if I run `:hi! link yamlKey DraculaCyan` manually it actually works! Yaml keys are colored in cyan as expected. I don't know why this does not work if I put the command in Dracula syntax file.

benknoble:
Interesting that it works manually. It shouldn't be an ordering issue, since `hi!` will overwrite, but `hi` will not. Hm.

TheWall89:
Hi,

sorry for coming back at this after a long time but I finally decided to give it another shot.

I have set up a clean VIM environment on a fresh Ubuntu VM. First, I colored the groups listed [here](https://github.com/dracula/vim/pull/163#issuecomment-568319158) (the correct ones). Then, I started playing around with syntax files in both dracula and polyglot and I found something. If I remove this `if` block
```
if dracula#should_abort('yaml')
    finish
endif
```
in `.vim/bundle/dracula/after/syntax/yaml.vim`, YAML files get finally colored.

image

I don't know why VIM enters in that `if` block while loading scripts. I'm attaching a log with V9.
[dracula-yaml.log](https://github.com/dracula/vim/files/4174086/dracula-yaml.log)

benknoble:
This is super odd to me: if I put `hi! link yamlKey DraculaCyan` in dracula's `after/syntax/yaml.vim`, the link works perfectly, but I do not have polyglot installed.

That if should still go through, unless stephpy is doing something bizarre...

TheWall89:
Yep! You are right! Those lines are the problem. If I remove them, the YAML file gets colored.

Should I open an issue to stephpy asking to remove those lines?

dsifford:
It's likely just the extra `syntax clear` after the condition that's the problem.. Look like the author was maybe debugging or something and forgot about it.

benknoble:
@TheWall89 yeah seems best.

TheWall89:
Issue open here: https://github.com/stephpy/vim-yaml/issues/22

sheerun:
I've removed stephpy/vim-yaml from vim-polyglot because it seems it causes issues and author is not responsive on github

benknoble:
@sheerun thanks for the followup!

PR #170, opened 2020-03-03
Neovim : add lsp message hightlighting

vigoux:
Neovim now supports LSP protocol, and this commit adds highlighting for
it.

Below is an example of what changes on Rust code using `rls`.
Before :
Capture d’écran 2020-03-03 à 09 29 48

After :
Capture d’écran 2020-03-03 à 09 30 23


dsifford:
LGTM, thanks!

PR #186, opened 2020-05-05
Added Lua syntax file

3uclidian:
Added Lua syntax file since built in functions weren't getting highlighted and
changed {} to green to differentiate from almost everything being pink.
(left: before, right: after)
screenshot

### DIFF
--- a/after/syntax/lua.vim
+++ b/after/syntax/lua.vim
@@ -2,5 +2,5 @@ if dracula#should_abort('lua')
     finish
 endif
 
-hi! link luaFunc DraculaCyan
-hi! link luaTable DraculaGreen
+hi! link luaFunc  DraculaCyan
+hi! link luaTable DraculaFg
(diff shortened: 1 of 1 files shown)


after/syntax/lua.vim
```diff
@@ -0,0 +1,6 @@
+if dracula#should_abort('lua')
+    finish
+endif
+
+hi! link luaFunc DraculaCyan
+hi! link luaTable DraculaGreen
```
dsifford, line 6:
```suggestion
hi! link luaFunc  Function
hi! link luaTable DraculaFg
```

3uclidian: In the spec it says [built ins should be cyan](https://spec.draculatheme.com/#sec-Language-Built-ins) unless I'm misreading it. Function declarations (with the function keyword) are green from the well named and completely distinct from "luaFunc", luaFunction highlight group.

PR #196, opened 2020-06-22
[update] airline's terminal section color

kazukazuinaina:
I've updated the color scheme in the terminal section of the vim-airline.
This is expected to improve the user interface.
This uses the same green color scheme as the insert mode color scheme.

### before

before

### after

after


benknoble:
:+1: LGTM @dsifford ?

dsifford:
Not a nvim terminal user, but seems reasonable to me.

kazukazuinaina:
Thank you for merging.
> Not a nvim terminal user, but seems reasonable to me.

This patch works fine in both Vim and Neovim.

PR #197, opened 2020-06-25
Make it possible to place dracula in the opt package directory.

cbtlsl:
This edit makes it possible to place dracula in the opt package directory by placing the creation of g:dracula_palette inside of a function that is called with the colorscheme is loaded.

### DIFF
--- a/autoload/dracula.vim
+++ b/autoload/dracula.vim
@@ -1,45 +1,47 @@
 " Palette: {{{
 
-let g:dracula#palette           = {}
-let g:dracula#palette.fg        = ['#F8F8F2', 253]
+func! dracula#create_palette ()
+  let g:dracula#palette           = {}
+  let g:dracula#palette.fg        = ['#F8F8F2', 253]
 
-let g:dracula#palette.bglighter = ['#424450', 238]
-let g:dracula#palette.bglight   = ['#343746', 237]
-let g:dracula#palette.bg        = ['#282A36', 236]
-let g:dracula#palette.bgdark    = ['#21222C', 235]
-let g:dracula#palette.bgdarker  = ['#191A21', 234]
+  let g:dracula#palette.bglighter = ['#424450', 238]
+  let g:dracula#palette.bglight   = ['#343746', 237]
+  let g:dracula#palette.bg        = ['#282A36', 236]
+  let g:dracula#palette.bgdark    = ['#21222C', 235]
+  let g:dracula#palette.bgdarker  = ['#191A21', 234]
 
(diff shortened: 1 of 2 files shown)


benknoble:
Thanks for the contribution @cbtlsl ! Can you explain how this makes using `opt` possible, and why it wasn't before ?

cbtlsl:
Files in the autoload directory are for functions that can be loaded automatically.  g:dracula_palette is not set in a function, so there is no guarantee that the variable will be created.  When the colorscheme is in an opt directory, packadd only adds the colorscheme directories to the runtimepath.  autoload/dracula.vim is not sourced and g:dracula_palette is not created.  So when you try to set the colorscheme you get a bunch of errors saying that the variable doesn't exist and the colors are not set correctly.

benknoble:
I remember some of the errors you're talking about, though it could be handy to have a reference (need to look in the old issues when I have time).

Also note that `:help autoload` says

```
This also works when reading a variable that has not been set yet: >

	:let l = foo#bar#lvar

However, when the autoload script was already loaded it won't be loaded again
for an unknown variable.
```

So it is unclear to me that the true cause of the errors is that `g:dracula#palette` is not set; by all accounts, it *should* trigger the use of the autoloaded file on lines like [this one](https://github.com/dracula/vim/blob/3916fb2daff546c6fe61d141b0e7519fde347b79/colors/dracula.vim#L34).

That is, the sequence should be

1. dracula is added to the runtimepath (via `packadd` or other mechanism)
2. `colorscheme dracula` triggers the linked line, which causes vim to do the equivalent of `runtime autoload/dracula` and then do the actual lookup on `g:dracula#palette`
3. ... the rest of the colorscheme is set up

P.S. I'm really trying to elicit the best solution to what is clearly a problem here; I just want to make sure we aren't fixing the wrong thing, so please don't take this as a criticism of the PR. I (we) really am (are) grateful for the contributions!

dsifford:
My thoughts mirror @benknoble -- I'm gonna be off the grid this weekend so once this is fleshed out, feel free to merge in @benknoble.

cbtlsl:
I have been doing a little more research and have found what is going on.  I have been using packadd! to add the package.  The '!' tells vim to only add the package to the runtime directory.  'packadd' without the '!' tells vim to add the package to the runtime directory and source the files.

So if I use packadd instead of packadd! everything works.

benknoble:
@cbtlsl wonderful new! Does that mean we don't need the PR?

benknoble:
I'm closing this since we haven't heard anything; feel free to re-open if we need to address something.

PR #198, opened 2020-07-02
JSX - Support vim-jsx-pretty and improve YATS syntax

### LINKED ISSUE #193: html tags color white in JSX syntax
### What happened
JSX syntax does not highlighting tags properly:
Screenshot from 2020-05-29 15-31-44@2x
### What I expected to happen
Tags should not be a white color

### Machine Info
- **Vim type (`vim`/`gvim`/`neovim`)**: Vim
- **Vim version**: 8.0.1453
- **OS**: Elementary OS
- **Terminal/Terminal Emulator/VTE**: tmux
- **`TERM` environment variable**: xterm-256color

Is that a bug? or it's normal that tags color white inside JSX syntax

renatoagds:
Closes #193 

#### SCREENSHOTS 

_vim-jsx-pretty default_

Screen Shot 2020-07-05 at 14 28 39

_with vim-jsx-pretty highlight close tag_

Screen Shot 2020-07-05 at 14 28 18

**EDIT**: Updated images matching correct colors.

### DIFF
--- a/after/syntax/javascript.vim
+++ b/after/syntax/javascript.vim
@@ -1,4 +1,4 @@
-if dracula#should_abort('javascript', 'javascriptreact', 'javascript.jsx') 
+if dracula#should_abort('javascript', 'javascriptreact', 'javascript.jsx')
   finish
 endif
 
@@ -19,8 +19,6 @@ hi! link jsDocTypeBrackets         DraculaCyan
 hi! link jsFuncArgOperator         Operator
 hi! link jsFunction                Keyword
 hi! link jsTemplateBraces          Special
-hi! link jsModuleAsterisk          DraculaPurpleItalic
-hi! link jsVariableDef             DraculaPurple
 
 "}}}
 
--- a/after/syntax/javascript.vim
+++ b/after/syntax/javascript.vim
@@ -20,6 +20,7 @@ hi! link jsFuncArgOperator         Operator
 hi! link jsFunction                Keyword
 hi! link jsTemplateBraces          Special
 hi! link jsModuleAsterisk          DraculaPurpleItalic
+hi! link jsVariableDef             DraculaPurple
 
 "}}}
 
--- a/after/syntax/javascript.vim
+++ b/after/syntax/javascript.vim
@@ -1,5 +1,5 @@
-if dracula#should_abort('javascript')
-    finish
+if dracula#should_abort('javascript', 'javascript.jsx') 
+  finish
 endif
 
 hi! link javaScriptBraces   Delimiter
@@ -19,7 +19,14 @@ hi! link jsDocTypeBrackets         DraculaCyan
 hi! link jsFuncArgOperator         Operator
 hi! link jsFunction                Keyword
 hi! link jsTemplateBraces          Special
+hi! link jsModuleAsterisk          DraculaPurpleItalic
 
 "}}}
+
+" maxmellon/vim-jsx-pretty 
+
+hi! link jsxComponentName   Type
+hi! link jsxAttrib          DraculaGreenItalic
+hi! link jsxCloseString     DraculaFg
 
 " vim: fdm=marker ts=2 sts=2 sw=2 fdl=0:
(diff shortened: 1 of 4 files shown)


dsifford:
Thanks @renatoagds 

Off the bat, I'm noticing the colors are not aligned with the spec and these changes negatively impact the highlighting for YATS.

If we can match how it was before for YATS and ideally translate that highlight to vim-jsx-pretty, then we should be good to merge.

No time until next week since I'm about to head out for vacation, but I think this should be able to be massaged a bit and worked out without an issue.

Here's an example of my YATS before and after...

**Before:**

image

**After:**

image

Essentially, the colors should follow the same format as we do for html/xml.

after/syntax/javascript.vim
```diff
@@ -19,7 +19,18 @@ hi! link jsDocTypeBrackets         DraculaCyan
 hi! link jsFuncArgOperator         Operator
 hi! link jsFunction                Keyword
 hi! link jsTemplateBraces          Special
+hi! link jsModuleAsterisk          DraculaPurpleItalic
+hi! link jsVariableDef             DraculaPurple
```
dsifford, line 23: Outside the scope of this PR. Please revert.

after/syntax/typescriptreact.vim
```diff
@@ -1,10 +1,27 @@
...
+if dracula#should_abort('typescriptreact', 'typescript.tsx')
     finish
 endif
 
 runtime! syntax/typescript.vim
 
-hi! link tsxAttrib           DraculaGreenItalic
-hi! link tsxEqual            Operator
-hi! link tsxIntrinsicTagName Keyword
-hi! link tsxTagName          Type
```
dsifford, line 10: These should have not been changed. Please revert.

after/syntax/javascript.vim
moQuez: Actually, those values were picked by me to match how it looked on VSCode. I'm ok whether we leave them or not.

dsifford: I'm the maintainer of the vscode repo. Variables _should_ be white by default. The reason why they appear to be purple in vscode is because vscode incorrectly treats `const` variables as true constants. There is an issue open in the repo to fix this.

The module asterisk thing is fine, but it's outside the scope of this particular PR. So I would just ask that it is opened in a different PR.

moQuez: Oh! Then I was trying to replicate a bug haha!... Thanks for spotting that out, and also thanks for your work on both repos then 🚀

after/syntax/javascript.vim
```diff
@@ -19,7 +19,18 @@ hi! link jsDocTypeBrackets         DraculaCyan
...
 
-"}}}
```
benknoble, line 23: This part of the diff is just noise from a commit that removed fold-markers where it shouldn’t have—it would be nice to restructure the patch series to avoid that, if possible. A `rebase -i` could do it, with some care (I suggest working in a separate branch that starts on the same tip; makes it easier to experiment). 

The idea being to avoid noise both in the history and in the blame... this makes it easier to gather actual information from the commit history of a file.

benknoble:
I’ve mostly stayed out of the way in all this because I’m not a heavy js(x)/ts(x) user. I left some general feedback on the commit structure—other than that, I trust @dsifford to guide this to the right state.

after/syntax/javascript.vim
```diff
@@ -1,5 +1,5 @@
-if dracula#should_abort('javascript')
-    finish
+if dracula#should_abort('javascript', 'javascriptreact', 'javascript.jsx')
```
benknoble, line 1: What if we fixed this function to check all current filetypes (split `&filetype` on a dot)? Then we could maybe not repeat ourselves and say 

```
if dracula#should_abort(..., 'jsx')
```

dsifford: I like this idea.

renatoagds:
@dsifford If you take a look at the print screens, we have two cases: `default` and `highlight close tag`, the second one is a `vim-jsx-pretty` config (`g:vim_jsx_pretty_highlight_close_tag`), that's why the `close tag` changes it color.

benknoble:
@dsifford Seems like there is strong interest in this—my thinking is if everyone’s content that this works, let’s close it out. We can open a new issue on the architectural side to improve the code there.

Issue #190, opened 2020-05-15
how to install dracula theme on spacevim?


spazmantiz:

benknoble:
I’m not a spacevim user, and I don’t think @dsifford is either. Does the spacevim project have documentation on how to install custom plugins or packages?

dsifford:
@benknoble https://spacevim.org/documentation/#custom-configuration

Looks like it's a dein wrapper under the hood.

benknoble:
Welp, looks like best thing to do is add dracula to those custom plugins?

spazmantiz:
> Welp, looks like best thing to do is add dracula to those custom plugins?

yes please, i want easy install

dsifford:
Follow instructions on this page: https://spacevim.org/layers/colorscheme/

spazmantiz:
i manage to install it by editing my **HOME/.Spacevim.d/init.toml** using this command
[[custom_plugins]]
  name = "dracula/vim"
[options]
  colorscheme = "dracula"

spazmantiz:
should i close it?

benknoble:
(Hope nobody minds, but I cleaned up the solution comment formatting)

spazmantiz:
i think you guys should also add guide to install dracula for vim distributions

PR #204, opened 2020-07-19
Update guide for installing dracula in spacevim

wsdjeg:
ref: https://github.com/dracula/vim/issues/190
ref: https://github.com/SpaceVim/SpaceVim/issues/3513


benknoble:
If it works, then I'm alright with it--I don't use SpaceVim

PR #212, opened 2020-10-18
Add RST syntax colors

d7xdev:
First PR, feedback and comment are welcomed. Thanks a lot!

- Add Dracula RST syntax colors to approximate Dracula Markdown syntax colors. RST is the other popular documentation format. Dracula RST colors is a nice feature to have when editing documentation between different projects. 
- Fix undefined `g:dracula_undercurl` when user disables underline. Current workaround requires the user to also disable undercurl when disabling underline. The fix is minor, but improves UX with least surprise.

### DIFF
--- a/after/syntax/rst.vim
+++ b/after/syntax/rst.vim
@@ -1,0 +1,26 @@
+if dracula#should_abort('rst')
+    finish
+endif
+
+hi! link rstComment                             DraculaComment
+hi! link rstCitationReference                   DraculaCyan
+hi! link rstFootnoteReference                   DraculaCyan
+hi! link rstHyperLinkReference                  DraculaCyan
+hi! link rstHyperlinkTarget                     DraculaCyan
+hi! link rstInlineInternalTargets               DraculaCyan
+hi! link rstInterpretedTextOrHyperlinkReference DraculaCyan
+hi! link rstCodeBlock                           DraculaGreen
+hi! link rstInlineLiteral                       DraculaGreen
+hi! link rstLiteralBlock                        DraculaGreen
+hi! link rstQuotedLiteralBlock                  DraculaGreen
+hi! link rstStandaloneHyperlink                 DraculaLink
+hi! link rstStrongEmphasis                      DraculaOrangeBold
+hi! link rstDirective                           DraculaPink
+hi! link rstSubstitutionDefinition              DraculaPink
+hi! link rstSections                            DraculaPurpleBold
+hi! link rstTransition                          DraculaPurpleBold
+hi! link rstTodo                                DraculaTodo
+hi! link rstCitation                            DraculaYellow
+hi! link rstExDirective                         DraculaYellow
+hi! link rstFootnote                            DraculaYellow
+hi! link rstEmphasis                            DraculaYellowItalic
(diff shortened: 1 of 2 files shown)


benknoble:
Can you split this into two PRs, please? One for E121, which we should be able to merge right away, and one for syntax/rst, which *may* require more discussion.

d7xdev:
Sure. I don't know how to edit the existing PR, but I can re-submit as two PRs.

after/syntax/rst.vim
```diff
@@ -0,0 +1,26 @@
+if dracula#should_abort('rst')
+    finish
+endif
+
+hi! link rstComment                             DraculaComment
```
dsifford, line 5: Please use built-in highlight groups where able.

```suggestion
hi! link rstComment                             Comment
```

dsifford (changes requested):
Looks good for the most part.

However, just one suggestion: Can you use built-in highlight groups where you're able to rather than referencing the dracula groups?

See the one example I provided inline...

after/syntax/rst.vim
d7xdev: No problem. Sounds much better. I'll do the correction.

benknoble: FWIW, I suspect @dsifford was referring to those that have obvious mappings (`Comment`, `Todo`), but not all of them do (why use `String` or `Keyword`, for example?).

But if it makes sense to you all, works for me.

after/syntax/rst.vim
```diff
@@ -0,0 +1,26 @@
...
+hi! link rstCodeBlock                           DraculaGreen
+hi! link rstInlineLiteral                       DraculaGreen
+hi! link rstLiteralBlock                        DraculaGreen
+hi! link rstQuotedLiteralBlock                  DraculaGreen
+hi! link rstStandaloneHyperlink                 DraculaLink
+hi! link rstStrongEmphasis                      DraculaOrangeBold
+hi! link rstSections                            DraculaPurpleBold
+hi! link rstEmphasis                            DraculaYellowItalic
+hi! link rstDirective                           Keyword
+hi! link rstSubstitutionDefinition              Keyword
```
d7xdev, line 16: I was not sure where to put RST Directive and Substitution Definition. I did consider `Label`, `Macro`, `Typedef` and `Define`. At the end, I choose `Keyword` because how significant Directive and Substitution are in RST.

after/syntax/rst.vim
d7xdev: I am sorry for the misunderstanding. I did more work then needed. But now that is done, I kinda like `Keyword` and `String`. This makes it easier for user to override these colors later.

benknoble: Given that the things you linked to Tag seem to be link-related, I would say that makes at least semantic sense to me.

d7xdev: Thanks for your insight. I dont mind hacks in my personal Vim config, but I dont want add burden on future maintenance. I am glad it's not too bad!

d7xdev:
I am sorry I may have corrupted my working branch. I'll close this PR, cleanup and resubmit a new one. Thanks again for the feedback and comments. Cheers!

PR #213, opened 2020-10-19
Fix E121: Undefined variable `g:dracula_undercurl`

d7xdev:
The variable `g:dracula_undercurl` is not properly defined
when underline attribute is disabled by user configuration:

    let g:dracula_underline = 0

The fix always defines missing `g:dracula_undercurl`,
but also disables it when underline is disabled.

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -82,11 +82,7 @@ if !exists('g:dracula_underline')
 endif
 
 if !exists('g:dracula_undercurl')
-  if g:dracula_underline == 0
-    let g:dracula_undercurl = 0
-  else
-    let g:dracula_undercurl = 1
-  endif
+  let g:dracula_undercurl = g:dracula_underline
 endif
 
 if !exists('g:dracula_inverse')
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -81,8 +81,12 @@ if !exists('g:dracula_underline')
   let g:dracula_underline = 1
 endif
 
-if !exists('g:dracula_undercurl') && g:dracula_underline != 0
-  let g:dracula_undercurl = 1
+if !exists('g:dracula_undercurl')
+  if g:dracula_underline == 0
+    let g:dracula_undercurl = 0
+  else
+    let g:dracula_undercurl = 1
+  endif
 endif
 
 if !exists('g:dracula_inverse')


colors/dracula.vim
```diff
@@ -81,8 +81,12 @@ if !exists('g:dracula_underline')
...
 
-if !exists('g:dracula_undercurl') && g:dracula_underline != 0
-  let g:dracula_undercurl = 1
+if !exists('g:dracula_undercurl')
+  if g:dracula_underline == 0
+    let g:dracula_undercurl = 0
+  else
+    let g:dracula_undercurl = 1
+  endif
 endif
```
dsifford, line 90:
```suggestion
if !exists('g:dracula_undercurl')
  let g:dracula_undercurl = g:dracula_underline == 0 ? 0 : 1
endif
```

Looks good. I prefer the ternary style though.

dsifford (changes requested):

colors/dracula.vim
benknoble: I would suggest not comparing with 0—there are lots of other "falsey" options:

```
let g:dracula_undercurl = g:dracula_underline ? 1 : 0
```

(or `!g:dracula_underline ? 0 : 1`)

I do not suggest doing `!!g:dracula_underline`--it works, but is hard to read.

benknoble: Not that people do this, but `""` and `"0"` are both falsey, so `let g:dracula_underline = ""` is doable (although, by the magic of vimscript, `"" == 0` is truthy 🙄 but `"" is# 0` is falsey ).

More likely is `v:false`, which I use these days when I can. (Yet again it `==` 0, but *not* `is#` 0.)

I guess my main issues are
1. Comparing boolean to a literal in a condition (better to just the boolean you mean), and
2. It may not be obvious that `==` coerces lots of possible values to 0 correctly, so it isn't clear that other values are valid.

It also means if we ever switch from `==` to the safer `is#`, we will actually introduce a regression (some things `== 0`, `isnot# 0`, and yet are still falsey because of `==`).

TL;DR, booleans in vim are complex :)

benknoble: Heck, the only reason to write the ternary is to force a known value of 1 or 0. We might as well right
```
let g:dracula_undercurl = g:dracula_underline
```
because it has the same logical meaning.

colors/dracula.vim
```diff
@@ -81,8 +81,8 @@ if !exists('g:dracula_underline')
   let g:dracula_underline = 1
 endif
 
-if !exists('g:dracula_undercurl') && g:dracula_underline != 0
-  let g:dracula_undercurl = 1
+if !exists('g:dracula_undercurl')
+  let g:dracula_undercurl = g:dracula_underline
```
benknoble, line 85: LGTM. @dsifford ?

benknoble (approved):

colors/dracula.vim
dsifford: Yep!

d7xdev: Thanks for your review and explanation. Cheers!

PR #215, opened 2020-10-31
add highlights for tbastos/vim-lua

nanotee:
Hello!

The default Vim syntax file for Lua is somewhat lacking, so I added some highlights for [tbastos/vim-lua](https://github.com/tbastos/vim-lua) (which is bundled with [vim-polyglot](https://github.com/sheerun/vim-polyglot/)

I tried mapping the highlight groups in accordance with the [spec](https://spec.draculatheme.com/):

- `luaBraces` -> `BracketsParensBraces`
- `luaBuiltIn` -> `InstanceReservedWords`
- `luaDocTag` -> `DocCommentKeywords`
- `luaErrHand` -> `BuiltInFunctions`
- `luaFuncArgName` -> `FunctionParameters`
- `luaFuncCall` -> `FunctionNames`
- `luaLocal` -> `Keyword`
- `luaSpecialTable` -> `BuiltInMagicMethodsOrConstants`
- `luaSpecialValue` -> `BuiltInFunctions`

Here's how it looks compared to the reference implementation:

screenshots

Thanks for maintaining `dracula/vim`!

### DIFF
--- a/after/syntax/lua.vim
+++ b/after/syntax/lua.vim
@@ -8,7 +8,7 @@ hi! link luaTable DraculaFg
 " tbastos/vim-lua {{{
 
 hi! link luaBraces       DraculaFg
-hi! link luaBuiltIn      DraculaPurpleItalic
+hi! link luaBuiltIn      Constant
 hi! link luaDocTag       Keyword
 hi! link luaErrHand      DraculaCyan
 hi! link luaFuncArgName  DraculaOrangeItalic
--- a/after/syntax/lua.vim
+++ b/after/syntax/lua.vim
@@ -14,7 +14,7 @@ hi! link luaErrHand      DraculaCyan
 hi! link luaFuncArgName  DraculaOrangeItalic
 hi! link luaFuncCall     Function
 hi! link luaLocal        Keyword
-hi! link luaSpecialTable DraculaPurple
+hi! link luaSpecialTable Constant
 hi! link luaSpecialValue DraculaCyan
 
 " }}}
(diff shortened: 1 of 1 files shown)


benknoble:
I'm not too familiar with that particular syntax highlighter for lua. But:

> `luaBuiltIn` -> `InstanceReservedWords`

Why not [https://spec.draculatheme.com/#BuiltInMagicMethodsOrConstants]() or [https://spec.draculatheme.com/#BuiltInFunctions]()?

I'm not saying you're wrong, but I need the rationale behind the decision (`Builtin` sounds like, well, Language Builtins, at least to me).

Other than that, assuming you used the write `Dracula*` highlight groups, LGTM.

dsifford (approved):
I'll leave the final merge of this to @benknoble pending the answer to his question.

Otherwise, LGTM. Really nice effort and nicely stuck to the convention, @nanotee. Good work!

nanotee:
`luaBuiltin` is used to highlight both `_ENV` and `self`: https://github.com/tbastos/vim-lua/blob/fa810f85437525bcea0fc4ff22c634935721d86d/syntax/lua.vim#L121

I opted for `InstanceReservedWords` because `self` seems to be a more common occurrence, but `BuiltInMagicMethodsOrConstants` could be valid as well

benknoble:
@nanotee I don't have a strong preference for what it should be. Why don't you try both and see what you like better?

nanotee:
@benknoble I initially preferred `InstanceReservedWords` because it's more consistent with the VSCode implementation for `self`, but I think your earlier suggestion is more future-proof. The name `luaBuiltIn` suggets that new keywords that aren't related to object orientation could be added in the future.

I have pushed the relevant changes, `luaBuiltIn` is now `BuiltInMagicMethodsOrConstants`

benknoble:
Alright, sounds good to me. Merging.

PR #217, opened 2020-11-13
fix(nvim): fix lsp hlgroups

vigoux:
A recent push to neovim master changed the highlight groups for lsp.
I know I am cheating here but the screenshots are exactly those over at https://github.com/dracula/vim/pull/170.


dsifford:
Should the old groups still be left in code for now for those who are not tracking the master branch of nvim?

vigoux:
LSP has not been released yet, so I think it is fine to actually change these.

dsifford:
Ah, okay then LGTM.

dsifford (approved):

PR #218, opened 2020-12-03
Syntax Highlighting specific for PureScript

andys8:
I started adding specific syntax for PureScript.

The intention was to avoid having the same color in the module and import line. The change for operator/backtick doesn't actually change the color, but is semantically right.

## Syntax Variables

The names are taken from [gruvbox](https://github.com/morhetz/gruvbox/blob/040138616bec342d5ea94d4db296f8ddca17007a/colors/gruvbox.vim#L1218-L1233
).

### Used

* purescriptModuleKeyword
* purescriptImportKeyword
* purescriptHidingKeyword
* purescriptAsKeyword
* purescriptOperator
* purescriptBacktick

### Unused

* purescriptStructure
* purescriptModuleName
* purescriptWhere
* purescriptDelimiter
* purescriptType
* purescriptTypeVar
* purescriptConstructor
* purescriptFunction
* purescriptConditional

## Screenshot

image

### DIFF
--- a/after/syntax/purescript.vim
+++ b/after/syntax/purescript.vim
@@ -1,0 +1,9 @@
+if dracula#should_abort('purescript')
+    finish
+endif
+
+hi! link purescriptModule Type
+hi! link purescriptImport DraculaCyan
+hi! link purescriptImportAs DraculaCyan
+hi! link purescriptOperator Operator
+hi! link purescriptBacktick Operator


dsifford:
Thanks @andys8!

Looks good so far. Low on time at the moment, but just quickly make sure the changes conform to the [spec](https://spec.draculatheme.com) as best as you can and either me or @benknoble will get this merged. :+1:

benknoble:
I haven't checked the spec matching. I was going to compare this to the Haskell highlighting, but the default Haskell highlighting must be pretty poor because it's not very good (I know not all of the names make sense, but I do know the syntaxes are similar enough to be comparable):

Capture d’écran 2020-12-05 à 13 48 00

So, sounds good to me. But if you have recommendations for Haskell syntax, or there are changes that would be good to make here, let me/us know!

benknoble:
Oof, looks like we dropped the ball on this one. Have we got any comparisons of this PR with the spec?

andys8:
Initially I thought it was a good idea to not have the same color for import lines (see screenshot). Not sure if this is okay with the spec.

What's the best to progress? Should Haskell be used for comparison?

benknoble:
No, I would stick with the spec; anything subjective we can discuss here (with screenshots, preferably).

andys8:
So, comparing with Haskell, Elm or Reason won't work, because there is these languages are all missing. I tried to start from scratch. I didn't understand the spec for my questions. So I did compare with typescript and tried to make it look similar.

## Before

image


## After

image

```
if dracula#should_abort('purescript')
    finish
endif

hi! link purescriptModule Type
hi! link purescriptImport DraculaYellow
hi! link purescriptImportAs DraculaCyan
hi! link purescriptOperator Operator
hi! link purescriptBacktick Operator
```

benknoble:
It does look a lot better… and the spec doesn't say much about imports, except at the end.

I am concerned about all that Yellow. It's a bit much for me. I would almost rather it be the Pink/Magenta color; honestly, I think I'm going to have to leave this to @dsifford to make the call. Improvements are great, but there's a fine line between works and works for everyone. And there's always the option to customize yourself (`:help dracula-customization`)

andys8:
Yeah, it looks yellow now, but that's what the reference (TypeScript) is using for whatever you are importing from.

I'm also fine with either way. Used PureScript and dracula long enough without any `syntax/purescript.vim` file, too ;)

benknoble:
pinging @dsifford again for review on this.

dsifford:
I don't have super strong feelings on this, but I do sort of tend to agree with @benknoble. The reason why the typescript imports are yellow is because the yellow part is literally just a string, which are yellow per the spec. 

The purescript imports are a bit muddier there because I assume they refer to a namespace or something?

Regardless, I don't personally use purescript so this change would have no impact on me personally. So if this is something that is preventing you from being productive, then I'd be fine with merging at least until other users complain about it.

andys8:
Oh, I see. Semantically in TypeScript this is a string to identify a module, and this might not be the best reference.

I'd then go with what `Haskell` is doing, which might be the better example.

#### Haskell
image

I then adapt it to use `cyan` for module names.

#### PureScript
image

dsifford:
Cool, let's go with that

dangtransontung:
How can I build 1 for .cpp and .c file?

benknoble:
@dangtransontung your question is not clear. One what? Did you mean to open new issue?

If you mean to customize C and C++ highlighting, you'll want to use `after/syntax/c.vim` and `after/syntax/cpp.vim`.

dangtransontung:
> @dangtransontung your question is not clear. One what? Did you mean to open new issue?
> 
> If you mean to customize C and C++ highlighting, you'll want to use `after/syntax/c.vim` and `after/syntax/cpp.vim`.

@benknoble Yes. that's what I mean because I cannot find c.vim and cpp.vim in after/syntax. Where should I find them or How can I write them? Do I need to purchase PRO edition to find the guide on writing them?

benknoble:
No need to purchase them. You can create them. You can either do it in your personal vim directory, or if you think it's useful to others, open a PR here. I recommend that if you do a PR here you follow the existing style in other such files.

Issue #231, opened 2021-03-08
Treesitter settings cannot be adjusted in init.vim


inventionate:
### What happened

I'm using the recently merged Treesitter support (#230). Thanks for that! Unfortunately, I can't change the colours in my init.vim. `highlight link TSError Normal` is unfortunately overwritten. I think this is especially annoying in the case of the live syntax errors. 

The problem has also been discussed elsewhere: https://github.com/nvim-treesitter/nvim-treesitter/issues/119

### What I expected to happen / possible solution

The solution is to use 'hi default link TSError ErrorMsg' instead of, for example, 'hi! link TSError ErrorMsg' (see https://github.com/nvim-treesitter/nvim-treesitter/pull/121). I would therefore advocate adapting all settings accordingly so that users can simply adapt them to their own needs.

### Screenshot

<img width="837" alt="Bildschirmfoto 2021-03-08 um 11 33 52" src="https://user-images.githubusercontent.com/872787/110310821-44a73c80-8003-11eb-94e5-3beb4923a66b.png">

### Machine Info

- **Vim type (`vim`/`gvim`/`neovim`)**: neovim
- **Vim version**: v0.5.0-dev+nightly-507-gc12ea02e0
- **OS**: Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64 x86_64
- **Terminal/Terminal Emulator/VTE**: iTerm2
- **`TERM` environment variable**: xterm-256color

benknoble:
Perhaps a better option is for your customization to be applied when the colorscheme changes: for that, use an autocommand.

```vim
augroup my_treesitter_overrides
  autocmd!
  autocmd ColorScheme dracula highlight …
augroup END
```

(Use `*` instead of `dracula` if you prefer.)

inventionate:
Thank you very much for your fast answer! Unfortunately, the autocommand solution does not seem to work. I cannot change TSError.

dsifford:
@inventionate can you paste your snippet that you tried?

benknoble:
I'm also not sure we should `:highlight default` those groups; that could cause unexpected issues elsewhere (e.g., other plugins)

inventionate:
I tried many different commands inside the augroup, e. g.:

```vim 
  autocmd ColorScheme dracula highlight TSError guibg=#282A36 guifg=#282A36
```
```vim 
[...]

PR #230, opened 2021-03-04
Add hightlight groups for nvim-treesitter

DerekStride:
# What

The neovim project [nvim-treesitter/nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter) ands some additional highlight groups for better syntax highlighting. I've adapted the work done by @nitishvelu in [ChristianChiarulli/nvcode-color-schemes.vim#14](https://github.com/ChristianChiarulli/nvcode-color-schemes.vim/pull/14) to support the treesitter hightlighting and would like to get it merged upstream in the main `dracula/vim` plugin.

## Rust

**Left**: Highlighting with nvim-treesitter

**Right**: Existing highlighting

Screen Shot 2021-03-04 at 13 03 32

## Ruby

**Left**: Highlighting with nvim-treesitter

**Right**: Existing highlighting

Screen Shot 2021-03-04 at 13 05 21

PR #232, opened 2021-03-08
Update treesitter highlight groups + remove syntax error highlights

DerekStride:
# What

I have 2 fixes for #230:

1) I removed duplicates when comparing with the treesitter defaults https://github.com/nvim-treesitter/nvim-treesitter/blob/master/plugin/nvim-treesitter.vim
2) I removed the highlight for `TSError`

I discovered that `TSError` is used for syntax errors from treesitter. These errors are very common, since treesitter is an incremental parser, it parses changes to the file regularly, and often in the middle of typing new code. This causes the invalid syntax to get highlighted with `TSError` as described in https://github.com/nvim-treesitter/nvim-treesitter/issues/78#issuecomment-647140700. More often than not we want to ignore invalid syntax while typing, so I've removed the highlight for `TSError`.

I don't think this will fix the root cause of https://github.com/dracula/vim/issues/231, however, I discovered this issue will attempting to debug it and this should provide a bandaid to the bad experience.

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -45,61 +45,31 @@ endif
 if exists('g:loaded_nvim_treesitter')
   " # Misc
   hi! link TSError ErrorMsg
-  hi! link TSPunctDelimiter Delimiter
-  hi! link TSPunctBracket Normal
   hi! link TSPunctSpecial Special
   " # Constants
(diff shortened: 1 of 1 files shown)


dsifford (approved):
LGTM. Thanks

PR #234, opened 2021-03-15
Add GitGutter/GitSigns highlighting

hahuang65:
This adds specific highlight groups for GitGutter-related things.
It piggybacks off of the `Diff*` highlight groups.

See these related links in from [gitsigns](https://github.com/lewis6991/gitsigns.nvim):
https://github.com/lewis6991/gitsigns.nvim/issues/54
https://github.com/lewis6991/gitsigns.nvim/commit/317750d66a572588eef9a23fefce4aff1cbcad94

The `gitsigns.nvim` maintainer [suggested](https://github.com/lewis6991/gitsigns.nvim/issues/94) that this be implemented by checking if neovim is >= version 0.5, and if the package is available (as opposed to a variable like `g:loaded_gitsigns`). This is because `gitsigns` is essentially lazy-loaded so checking the variable when Dracula loads might be too early.

Before:
Before

After:
After

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -41,6 +41,26 @@ if exists('g:loaded_ctrlp')
   hi! link CtrlPBufferHid Normal
 endif
 " }}}
+" GitGutter / gitsigns: {{{
+if exists('g:loaded_gitgutter')
+  hi! link GitGutterAdd    DiffAdd
+  hi! link GitGutterChange DiffChange
+  hi! link GitGutterDelete DiffDelete
+endif
+if has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
+  " https://github.com/lewis6991/gitsigns.nvim requires nvim > 0.5
+  " has('nvim-0.5') checks >= 0.5, so this should be future-proof.
+  hi! link GitSignsAdd      DiffAdd
+  hi! link GitSignsAddLn    DiffAdd
+  hi! link GitSignsAddNr    DiffAdd
+  hi! link GitSignsChange   DiffChange
+  hi! link GitSignsChangeLn DiffChange
+  hi! link GitSignsChangeNr DiffChange
+  hi! link GitSignsDelete   DiffDelete
+  hi! link GitSignsDeleteLn DiffDelete
+  hi! link GitSignsDeleteNr DiffDelete
+endif
+" }}}
 " Tree-sitter: {{{
 if exists('g:loaded_nvim_treesitter')
   " # Misc


benknoble:
General comments:

- the diff is too noisy, mostly because the lines with alignment don't substantially change (another argument in favor of not horizontally aligning everything)
- we've tried to keep plugin code out of the main `colors.vim`; could you move to somewhere in `after/plugin/dracula.vim`? Using a subfolder would be fine with me, too (e.g., `after/plugin/dracula/gitgutter.vim`)--@dsifford, this might be worth considering long-term.

hahuang65:
@benknoble essentially, the changes are only adding:

```vim
hi! link GitGutterAdd    DiffAdd
hi! link GitGutterChange DiffChange
hi! link GitGutterDelete DiffDelete
```

You're right the diff is noisy because of the alignment changes. What would you like me to do regarding this? I'm open since  you guys are the maintainers.

Regarding your second point, I don't mind moving out of the main `colors.vim` file, however, I would like to point out that this code is not specific plugin code. I'm not sure where it originated, but I know of at least a few plugins that use this highlight group. It possibly is a plugin only thing, I'm not sure, but it certainly isn't limited to a SINGLE plugin.

Knowing that (and maybe you already did), would you still like me to move it elsewhere?

dsifford:
@benknoble just click "ignore whitespace changes" in the github diff viewer.

Also, I use gitgutter personally and find that the default diff highlights look just fine. They do not look how @hahuang65 has in his screenshots so I'm curious if it's a configuration issue on their end.

hahuang65:
@dsifford I'm not actually using gitgutter. I'm using gitsigns. The maintainer of gitsigns mentioned it's because some themes treat DiffAdded and such as a background highlight.

I wonder if you tried gitsigns would it still be the same?

The curious thing to me is that... 1 week ago, it was working well. Then something changed and it looks like my screenshot too. I remember distinctly because I just moved to neovim and it was working on my Mac setup... Then when I cloned to setup to my Arch, I noticed it and thought it was the terminal emulator. But I came back to Mac and it's the same issue.

dsifford:
Interesting.

I grepped `GitGutterAdd` in my vim configs and it looks like it is indeed being handled by vim-gitgutter. gitsigns probably handles it slightly different so I'm not opposed to just adding in the explicit configuration if you want to just follow @benknoble's suggestions. Just make sure your `exists()` check accounts for both plugins.

benknoble:
> You're right the diff is noisy because of the alignment changes. What would you like me to do regarding this? I'm open since you guys are the maintainers.

Moving it out of `colors.vim` would fix this; there would be no re-alignment; hence no diff.

> Regarding your second point, I don't mind moving out of the main colors.vim file, however, I would like to point out that this code is not specific plugin code. I'm not sure where it originated, but I know of at least a few plugins that use this highlight group. It possibly is a plugin only thing, I'm not sure, but it certainly isn't limited to a SINGLE plugin.

`after/plugin/dracula.vim` is always sourced, so it doesn't matter that more than one plugin might care. (We do guard against doing things unnecessarily in there, though; in particular, `dracula#should_abort` and `exists()` checks.)

---

I looked at the linked conversation; gitsigns is irrelevant here. What is relevant is that you've asked gitsigns to use `GitGutter*` highlight groups. (Where those highlight groups get applied might be relevant, but that's a separate thing.)

At the end of the day, it comes down to either
- everyone agrees the new GitGutter groups look good, or
- we define separate groups for git signs use, or
- you make your own highlight groups, possibly out of the dracula-provided ones, when customizing your plugin setup (this has *always* been an option)

hahuang65:
@dsifford thanks for checking into that.

I just realized, I'm not sure I have an easy way to check if `gitsigns` is installed/running/available. It's a `nvim` only plugin, written in lua, so it doesn't define a `g:loaded_gitsigns` variable.

Any suggestions? Right now I've got:
```
if exists('g:loaded_gitgitter') || (has('nvim') && check_gitsigns)
```
where `check_gitsigns`, I don't know how to do.

dsifford:
Give this a try:

https://github.com/lewis6991/gitsigns.nvim/blob/a79dc28316b30ae0a8db203718c75e0a0f919564/doc/gitsigns.txt#L256

hahuang65:
@dsifford I took a look at that, and that might work, but I also filed a ticket with gitsigns to implement the `g:loaded_gitsigns` variable.

dsifford:
Excellent. Thanks for doing that.

dsifford:
Worst case scenario if all else fails we just don't do the check and add the two groups like you originally implemented.

hahuang65:
@dsifford yeah looks like `b:gitsigns_status` doesn't work, or maybe I need to do something to load `after/plugin/dracula.vim`?

I think worst case scenario, we should still implement it in `after/plugins/dracula.vim` and I can set `g:loaded_gitsigns` myself in my `vimrc` if gitsigns can't/won't implement it in their plugin.

It is cleaner for dracula this way, and we leave it up to users/plugins.

Now I'm thinking if there's a better way to be more inclusive to other git-related plugins than specifying plugin names individually? Maybe if `set signcolumn` is on? Is there a better way to tell if there's anything git-related that goes in to the sign column? Maybe I'm overthinking this.

hahuang65:
So, for some reason, if I add my 3 lines of change:

```
hi! link GitGutterAdd    DiffAdd
hi! link GitGutterChange DiffChange
hi! link GitGutterDelete DiffDelete
```
to `after/plugin/dracula.vim` it simply doesn't fix my issue. This is without any `if` statements, just the same 3 lines, but in a different file.

If I put those in `colors/dracula.vim`, it works perfectly.

Is there some sort of loading or order-of-operations thing I need to know about?

dsifford:
> Is there some sort of loading or order-of-operations thing I need to know about?

Yes, in this case these syntax groups must be defined before gitsigns loads because gitsigns uses these groups (which come from vim-gitgutter) as a fallback mechanism when they are defined.

hahuang65:
@dsifford if I'm understanding that correctly, that would imply that this entire effort is moot? If `gitsigns` is going to require `dracula` to be loaded first, then I cannot put a check to see if `gitsigns` is loaded... right?

dsifford:
Not by a computer so can't check right this moment but it wouldn't be moot if you instead target the gitsigns specific group (and not a fallback group from vim-gitgutter) assuming the plugin defines one.

Just a guess though. Can't say for sure without looking to confirm.

benknoble:
Perhaps another point in favor of creating your own highlight groups out of ours in your config, and using those for gitsigns…

hahuang65:
I just updated the PR with the code that I think will work if `gitsigns` doesn't implement a `g:loaded_gitsigns` variable.
It would be better if they do. That being said, perhaps @benknoble is right about just doing it in personal configs.

I don't know. It seems like it would be nice to fix it out-of-the-box for people using GitSigns.

benknoble:
The check for gitsigns will only fire when vim is starting up, so if the plugin is loaded but there are no signs placed yet (e.g., run `vim`, *then* open a file for there to be signs), the links won't be defined.

The problem seems to be that GitSigns doesn't make it easy to check—otherwise, I would be fine with the proposed change.

I would *consider* an un-guarded `highlight! link` for GitSigns while we wait on a fix from them, perhaps.

hahuang65:
Okay, I got a good answer from the maintainer of `gitsigns`. We should check if neovim is >= version 0.5, and if the package is available.

I've updated my code, as well as added a few more GitSigns-specific highlight groups.

The maintainer mentioned that they recently added their own GitSigns highlight groups, which is why Dracula was previously working for me, and then suddenly stopped highlighting correctly.

hahuang65:
I think this PR's code is in it's final state now. Please let me know your thoughts :) Thank you for all the time you guys have already spent on this, really appreciate it!

after/plugin/dracula.vim
```diff
@@ -2,6 +2,16 @@ if dracula#should_abort()
   finish
 endif
 
+function! s:GitSignsDefined()
+  " https://github.com/lewis6991/gitsigns.nvim requires nvim > 0.5
+  " has('nvim-0.5') checks >= 0.5, so this should be future-proof.
+  if has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
```
benknoble, line 8: Just return the boolean:

```vim
return has('nvim-0.5') && lueaeval("pcall(require, 'gitsigns')")
```

benknoble (changes requested):

benknoble:
Can you put a link to that in this PR or in one of the commit messages?

hahuang65:
Here is the [link to the issue](https://github.com/lewis6991/gitsigns.nvim/issues/94), just for anyone reading this in the future to follow.

I'll also put this in the description of the PR, as well as in the commit message.

hahuang65:
@benknoble updated! Thanks to you and @dsifford for sticking with me on this one. Really appreciate the help and guidance!

benknoble (approved):
I approve 🎉

after/plugin/dracula.vim
```diff
@@ -2,6 +2,12 @@ if dracula#should_abort()
   finish
 endif
 
+function! s:GitSignsInstalled()
+  " https://github.com/lewis6991/gitsigns.nvim requires nvim > 0.5
+  " has('nvim-0.5') checks >= 0.5, so this should be future-proof.
+  return has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
```
dsifford, line 8: I feel like this should just be inlined on line 56.

dsifford: @benknoble Don't have strong feelings on this, just a suggestion. I'm cool with you merging whenever you're ready. The reins are all you on this one.

hahuang65: I agree w/ the inlining. Changed the code.

dsifford:
Thanks for sticking it out @hahuang65!

benknoble:
Seconded; this was a tough one, but I definitely appreciate the improvements made.

hahuang65:
Thank you guys again for the help and guidance! It was a fun experience!

PR #241, opened 2021-04-07
fix(nvim): Add LspReference* groups

chasinglogic:
Adds LspReference groups so that the reference highlighting feature
works for neovim built-in LSP.

Before:

Screenshot 2021-04-07 at 19 29 44

After:

Screenshot 2021-04-07 at 19 30 28

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -250,6 +250,9 @@ call s:h('Conceal', s:cyan, s:none)
 " Neovim uses SpecialKey for escape characters only. Vim uses it for that, plus whitespace.
 if has('nvim')
   hi! link SpecialKey DraculaRed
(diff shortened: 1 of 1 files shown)


chasinglogic:
This closes #240

benknoble:
This looks like a net win to me; I don't have any objections. I also don't use neovim, so I can't confirm anything there (thanks for the screenshots). @dsifford ?

dsifford:
Looks good to me.. But what's going on with the line numbers in your screenshot? That's not related, correct?

chasinglogic:
@dsifford That's correct, I prefer the dracula line numbers from Emacs which have that background color so I have a customisation in my init.vim to change the `LineNr` highlight group.

Issue #244, opened 2021-04-11
Need better installation instructions for Windows


benknoble:
See [this QA](https://stackoverflow.com/q/67015374/4400820) with a response from @romainl, who also points out a potential issue with our `packadd!` instructions. (They are necessary because of #138 #139 #140 #142.)

romainl:
So you managed to drag me into this after all…

You should move your palette variables from `autoload/dracula.vim` to `colors/dracula.vim`.

By doing that, you:

- remove the dependency to `autoload/dracula.vim` from `colors/dracula.vim`,
- so you don't need to have  `dracula` in `&rtp`,
- so you don't need `packadd! dracula` in your `vimrc`,
- and you don't break the Airline and Lightline themes because the variables are defined long before they are sourced.

benknoble:
Sorry romainl; if you're not interested, I suppose you can unsubscribe. The airline issue arose when someone installed the colorscheme but did not activate it and attempted to activate the corresponding airline theme.

Do you have a better suggestion to avoid the duplication that doesn't involve the runtimepath? I suppose some `<sfile>:h` trickery might work.

PR #251, opened 2021-05-19
update installation instructions for other OS

### LINKED ISSUE #224: Prompt color in terminal inside vim changes
### What happened
When the colorscheme is active and I launch a terminal inside vim (via :terminal), the colors of the bash prompt are different from what I have set in .bashrc. Sometimes it changes them to the dracula colors (pink and yellow) and sometimes it is just plain white.

### What I expected to happen
The shell prompt inside vim should not differ from the one in any other terminal (emulator).

#### My standard bash prompt (outside vim):
grafik
#### The shell promt in a terminal inside vim:
grafik
grafik

### Machine Info
- **Vim type (`vim`/`gvim`/`neovim`)**: vim
- **Vim version**: 8.2 including patches 1-814
- **OS**: Linux gentoo 5.10.7-gentoo #1 SMP Fri Jan 15 21:48:43 CET 2021 x86_64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz GenuineIntel GNU/Linux
- **Terminal/Terminal Emulator/VTE**: alacritty
- **`TERM` environment variable**: alacritty
I have also tried setting the $TERM variable to diferent values like xterm-256color or tmux-256color. The results are the same.
### Additional Info
The same thing happens on my Artix system, so the problem also occurs on precompiled versions of vim and is not specific to my self-compiled vim on gentoo.

benknoble:
closes #224

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -35,7 +35,7 @@ git clone https://github.com/dracula/vim.git dracula
 ```
 (Or use your favorite GUI client, or download the ZIP)
 
-5. Edit your `vimrc` file with the following content:
+4. Edit your `vimrc` file with the following content:
 
 ```
 packadd! dracula
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,32 +7,35 @@ sections below, if you use other plugin managers.
 
 1. Create theme folder (in case you don't have yet):
 
+
+- \*nix:
 ```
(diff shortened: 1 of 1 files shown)


dsifford (approved):
LGTM. Nice! :+1:

benknoble:
Got the issue number wrong *sigh* should be #244

Issue #248, opened 2021-05-07
Strange pink highlight in some words - is this by design?


gregth:
### What happened
Some words in vim appear highlighted with pink color. For example, when using Tagbar plugin for `go` language outline, or at the intro screen of VIM. 

### What I expected to happen
This highlight is irritating, is this a bug or just a design decision?

### Screenshot
At right bottom is the Tagbar with the pink highlights:
![image](https://user-images.githubusercontent.com/8137385/117427645-eaa4f380-af2d-11eb-8ebc-e05c9e23ff6b.png)

Another case, at the vim intro screen:
![image](https://user-images.githubusercontent.com/8137385/117428675-f47b2680-af2e-11eb-8772-5fd4bd299468.png)

The following screenshot is what I get when I run vim without loading plugins, which means with Dracula for vim disabled (still I have Dracula theme for my gnome-terminal enabled): 

![image](https://user-images.githubusercontent.com/8137385/117427938-39eb2400-af2e-11eb-8f73-f8ba94b30411.png)

### Machine Info
Linux Dellis 4.15.0-142-generic #146-Ubuntu SMP Tue Apr 13 01:11:19 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

- **Vim type (`vim`/`gvim`/`neovim`)**: vim
- **Vim version**: VIM - Vi IMproved 8.0
- **OS**: Ubuntu 18.04
- **Terminal/Terminal Emulator/VTE**: gnome-terminal
- **`TERM` environment variable**: screen

benknoble:
Hmm, looks like the `SpecialKey` might be off--do you change that in your config? Could another plugin? Try `:verbose highlight SpecialKey`

benknoble:
Unfortunately for the moment, this is "intended"… due to a quirk (I'll try to dig up issue references) that neovim & vim use (slightly) different highlight groups for trailing whitespace, which we would like to be more subtle. Vim uses SpecialKey, which *also* get's used for stuff like `<CR>`/etc. (the output of `:map` is also not-fun if you have special keys). NeoVim, IIUC, uses a different group, so we can style the trailing whitespace separately from the special keys. I *think* @dsifford uses NeoVim and can give more context on this.

If you really don't like it, one simple fix is
```vim
augroup dracula_custom
  autocmd!
  autocmd ColorScheme dracula highlight link SpecialKey DraculaPink
augroup END
```
(or whatever your preferred color is)

dsifford:
[...]

Issue #247, opened 2021-04-28
Not able to see the NonText


ddaza:
### What happened
I have setup my NonText items to show in my config
```
set listchars=tab:❘·,trail:·,nbsp:·,precedes:«,extends:»,eol:↵
```
And I am unable to see these when I have both a cursor line and non text.

### What I expected to happen
Expected to show something like this. 
<img width="278" alt="Screen Shot 2021-04-28 at 10 01 45 AM" src="https://user-images.githubusercontent.com/290254/116427639-cf3d3880-a809-11eb-8703-f8befa86ab4e.png">

Showing this
<img width="277" alt="Screen Shot 2021-04-28 at 10 02 56 AM" src="https://user-images.githubusercontent.com/290254/116427705-dfedae80-a809-11eb-83b2-80e2c757583d.png">

### Machine Info

- **Vim type (`vim`/`gvim`/`neovim`)**: NVIM
- **Vim version**: v0.4.4
- **OS**: osx
- **Terminal/Terminal Emulator/VTE**: alacritty
- **`TERM` environment variable**: xterm-256color

benknoble:
Perhaps the cursorline is the same color and therefore hiding the symbols? if you move the cursor, do they appear? This is true for me in alacritty with a custom terminfo and termguicolors (I've actually stopped using cursorline).

(There was an issue somewhere else about something similar, maybe? Part of the issue is that vim uses maybe one or two colors for the `listchars` stuff, while neovim uses slightly different ones, maybe more, making it hard to get consistent. e.g., IIRC, @dsifford uses neovim and his trailing whitespace shows up as red?)

ddaza:
They never appear when I have my cursor line ON and on a line. 
And you are correct, the color of the CursorLine and the NonText is the same, so I think that is the culprit. 
I had to manually change the colors to make them appear.

DraculaSubtle and CursorLine are the same color. 

https://github.com/dracula/vim/blob/d82b9198d4dda1ac4a96756570f56125a1f86cb1/colors/dracula.vim#L150
https://github.com/dracula/vim/blob/d82b9198d4dda1ac4a96756570f56125a1f86cb1/colors/dracula.vim#L226
https://github.com/dracula/vim/blob/d82b9198d4dda1ac4a96756570f56125a1f86cb1/colors/dracula.vim#L208-L211

ddaza:
these were my changes to make it show. 
```diff
diff --git a/autoload/dracula.vim b/autoload/dracula.vim
[...]

Issue #210, opened 2020-10-06
digraphs hard to read


dubst3pp4:
### What happened

When using `:digraphs` the listed special characters are in dark grey and thus very hard to read.

### What I expected to happen

The usage of a more lighter color for the special characters.

### Screenshot

![Bildschirmfoto von 2020-10-06 11-00-53](https://user-images.githubusercontent.com/2373731/95181326-6d60a700-07c3-11eb-9d95-b3086d04cbfd.png)

### Machine Info

- **Vim type (`vim`/`gvim`/`neovim`)**: gvim
- **Vim version**: 8.2 Patch 1-1767
- **OS**: Linux 5.4.0-48-generic 52-Ubuntu SMP Thu Sep 10 10:58:49 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

dubst3pp4:
I think the syntax group is `SpecialKey` - I had the same issue with another color scheme (see https://github.com/haishanh/night-owl.vim/commit/18b6261656e6909ea93026de7eecdbe68686c3a9).

benknoble:
I agree this is hard to read; same issue exists for the output of `:map` family, IIRC. 

Open to suggestions... @dsifford ?

dsifford:
These show up red for me...

![image](https://user-images.githubusercontent.com/5240018/95208841-a266fc80-07b7-11eb-9ed7-059fc5d7f14c.png)

dubst3pp4:
@dsifford that's strange. When I execute `:highlight SpecialKey` I get the following:
```
SpecialKey     xxx term=bold ctermfg=81 guifg=Cyan
                   links to DraculaSubtle
```
I don't know what this means exactly...

dsifford:
I'm using neovim. Not sure if they treat digraphs differently.

benknoble:
@dsifford they do; we saw this come up once before I think.

`colors/dracula.vim` roughly line 250:
```
" Neovim uses SpecialKey for escape characters only. Vim uses it for that, plus whitespace.
if has('nvim')
  hi! link SpecialKey DraculaRed
  hi! link LspDiagnosticsUnderline DraculaFgUnderline
  hi! link LspDiagnosticsInformation DraculaCyan
  hi! link LspDiagnosticsHint DraculaCyan
  hi! link LspDiagnosticsError DraculaError
  hi! link LspDiagnosticsWarning DraculaOrange
  hi! link LspDiagnosticsUnderlineError DraculaErrorLine
[...]

Issue #192, opened 2020-05-29
listchars not visible when cursorline is enabled.


carlogravador:
### What happened
listchars(spaces, tabs) on a line are not visible when cursorline is active on that line. Also When the cursor is pointing on a listchar, it also becomes not visible

### What I expected to happen
Is it possible to change the colors of listchars and invert the color when cursorline or the cursor is on that position

### Screenshot
In this screenshot, on the 1st line, trailing spaces is visible and the colors are perfect for the theme
![image](https://user-images.githubusercontent.com/29529072/83220090-a6996280-a1a4-11ea-8c59-6ccd794fd9ec.png)
But on the second line, my cursor is on the first column of the line, but is not visible, also the trailing spaces are not visible because it blends with the cursorline color. I,m using indentline plugin to show indents.

### Machine Info
Linux probook 5.4.0-31-generic #35-Ubuntu SMP Thu May 7 20:20:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

- **Vim type**:NeoVim
- **Vim version**: v0.5.0-529-g8c588246a
- **OS**: Ubuntu 20.04
- **Terminal/Terminal Emulator/VTE**:GNOME Terminal 3.36.1.1
- **`TERM` environment variable**: xterm-256color

benknoble:
I don't know about solving this, but this is because `listchars` uses `SpecialKey` for trailing whitespace, which links to `DraculaSubtle` (`ctermfg=60`). `CursorLine` has `ctermbg=60`, so they blend.

As I said, I don't know about an official solution to this, but you could always do

```
autocmd ColorScheme dracula hi! CursorLine ...
```

to use your own `CursorLine` color

carlogravador:
Thank's for the answer, I've managed to achieve what I want by changing the colors of listchars.

```vimscript
highlight SpecialKey ctermfg=... guifg=...
highlight Whitespace ctermfg=... guifg=...
```
![image](https://user-images.githubusercontent.com/29529072/83324322-1851d900-a297-11ea-9a0b-56523b8cc6d7.png)

benknoble:
Could you share the value you used? I'm wondering if we should consider a change on this (I for one wouldn't be opposed to some of these characters being a little more visible; the output in `:map` with special characters for example is a bit hard to read at times).

PR #252, opened 2021-05-20
change SpecialKey to DraculaPink (vim only)

benknoble:
@dsifford this is my attempt to fix some of the listchars issues for vim. It does make some parts of `'listchars'` pink:

```
"SpecialKey" for "nbsp", "space", "tab" and "trail".
```

---

Related: #210, #248

Also #192 is about listchars + cursorline in NeoVim, which this doesn't target
Also #247 is about `NonText`, which needs investigating.

Issues: `:digraphs`, special keys in `:map` output, and certain parts of
`'listchars'` are "subtle" and hard to read (esp. w/ `cursorline` over
them). Also, vim uses `SpecialKey` for them (instead of `Whitespace` for
`'listchars'` and `SpecialKey` elsewhere, like NeoVim).

Fix: make `SpecialKey` more like special keys (the highlight in vimscript
for, e.g., `map <Space> :echo "hello"<CR>`). This has the side effect of
making digraphs more visible. It also makes certain `'listchars'` pink,
which may be annoying. But there is no way currently to separate the two
in vim.

Rationale: It matches with the special keys in maps. The fact that vim
uses it elsewhere may simply be "wrong" and need changed.

If you don't like it: well, as usual, you can use an autocommand to use
whatever colors you want (`:help dracula-customization`).

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -262,7 +262,7 @@ if has('nvim')
   hi! link LspDiagnosticsUnderlineInformation DraculaInfoLine
   hi! link LspDiagnosticsUnderlineWarning DraculaWarnLine
 else
-  hi! link SpecialKey DraculaSubtle
+  hi! link SpecialKey DraculaPink
 endif
 
 hi! link Comment DraculaComment


dsifford (approved):
Non-impactful change for me personally since I do indeed `has('nvim')` :stuck_out_tongue_closed_eyes: 

If this solves the issue on your side, then LGTM!

benknoble:
@dsifford I'm going to let this sit for a week before merging and try to gather feedback from at least those who had submitted issues.

dubst3pp4:
@benknoble Just checked out your branch with Vim 8.2 and digraphs are looking really good for me:
Bildschirmfoto von 2021-05-25 11-08-54
Thanks for your work!

benknoble:
@dubst3pp4 strangely enough they don't look pink, but you're welcome!

dubst3pp4:
I'm so sorry. I've used the wrong colorscheme. Now with dracula... looks very pink now ;-)

Bildschirmfoto von 2021-05-31 08-43-08

benknoble:
No worries @dubst3pp4 !

benknoble:
@dsifford it's been two weeks and I've not heard any negative feedback, at least from those I was able to request it from. If you're still go, let's ship it 🚀

dubst3pp4:
@benknoble for me it works great! Thank you very much!

gregth:
Thank you for the fix! It works great for me too!

ntessore:
This is an annoying change if one is using listchars to highlight spaces, tabs, etc.

For what it's worth, here is how I got it back to before, as per the commit message:

```vim
augroup dracula_customization
    au!
    autocmd ColorScheme dracula hi! link SpecialKey DraculaSubtle
augroup END
```

benknoble:
@ntessore yes, it does unfortunately make trailing spaces and other `listchars` features stand out more than they used to. We need vim to add support for a `Whitespace` group instead of re-using `SpecialKey` for those in `listchars`. In the meantime, yes your fix will go back to the old behavior.

Per48edjes:
Hi y'all, apologies if I'm missing something obvious here -- been using Dracula to teach myself a little vimscript (read: thanks for doing a solid job maintaining this project!), but I can't seem to get the autocommand group to make the desired changes (see below for a snippet from my `.vimrc`).

```vim
" Set color scheme
let g:dracula_colorterm = 0
colorscheme dracula
set cursorline

augroup dracula_customization
  au!
  autocmd ColorScheme dracula highlight! link SpecialKey DraculaSubtle
  autocmd ColorScheme dracula hi CursorLine cterm=underline term=underline ctermbg=NONE
augroup END
```
When I run `hi Special Key` after writing and sourcing (or closing and opening) I still see the following:

```vim
SpecialKey     xxx term=bold ctermfg=81 guifg=Cyan
                   links to DraculaPink
```

Same story re: the cursorline. I suspect it's some sort of rookie mistake, but having a hard time putting my finger on it. Any guidance would be helpful!

benknoble:
@Per48edjes looks like you figured this out, but yes, the autocommand needs to be installed before the command that triggers it (`colorscheme`, in this case).

PR #255, opened 2021-07-18
Highlight tag attributes via tree-sitter.

adriantrunzo:
This pull request sets the [correct color](https://spec.draculatheme.com/#HtmlCssAttributeNames) (italic green) for the  `TSTagAttribute` highlight group used by treesitter for tag attributes in JSX. By default, the nvim-treesitter package links `TSTagAttribute` to `TSProperty`, which in turns links to `Identifier`, which makes the attributes appear as white text.

Before:
Screen Shot 2021-07-17 at 7 35 46 PM

After:
Screen Shot 2021-07-17 at 7 24 24 PM

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -89,6 +89,8 @@ if exists('g:loaded_nvim_treesitter')
   hi! link TSTitle DraculaYellow
   hi! link TSLiteral DraculaYellow
   hi! link TSURI DraculaYellow
+  " # Tags
+  hi! link TSTagAttribute DraculaGreenItalic
 endif
 " }}}
 
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -62,6 +62,11 @@ if has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
 endif
 " }}}
 " Tree-sitter: {{{
+" The nvim-treesitter library defines many global highlight groups that are
+" linked to the regular vim syntax highlight groups. We only need to redefine
+" those highlight groups when the defaults do not match the dracula
+" specification.
+" https://github.com/nvim-treesitter/nvim-treesitter/blob/master/plugin/nvim-treesitter.vim
 if exists('g:loaded_nvim_treesitter')
   " # Misc
   hi! link TSPunctSpecial Special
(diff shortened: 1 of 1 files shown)


benknoble:
Is this highlight group defined globally or only for certain files (such as JS)?

If it affects more than JS, I want to know what other places it can appear in.

Otherwise LGTM

adriantrunzo:
> Is this highlight group defined globally or only for certain files (such as JS)?

Great question. It is defined globally. As far as I know all of the highlight groups in nvim-treesitter are global, even if they may only be used by a few languages.

> If it affects more than JS, I want to know what other places it can appear in.

The two places I find the group being applied are for [jsx](https://github.com/nvim-treesitter/nvim-treesitter/blob/fb5d6e04a87863d01a45dab2cbafec4e2a74822b/queries/jsx/highlights.scm#L7) and [html tags](https://github.com/nvim-treesitter/nvim-treesitter/blob/fb5d6e04a87863d01a45dab2cbafec4e2a74822b/queries/html_tags/highlights.scm#L4). In both instances it looks like the group was added just 5 days ago.. 

It is difficult to say how many small pull requests like this one will come up in the future. I honestly had no idea how tree-sitter/nvim-treesitter worked until doing a bit of research for that other issue yesterday, but I will keep an eye on it.

benknoble:
The global definition is odd, to me, but if it's only used in those places then I approve. @dsifford ?

benknoble (approved):

benknoble:
Actually, having said that, it might be nice to document (or at least link this PR and related issue in a comment) what the thing is for. `Tags` is not really an explanation (referring to the comment).

adriantrunzo:
@benknoble Good idea, I added some more comments. Let me know what you think.

dsifford:
I missed this somehow so my apologies. I'll kick the tires on this tomorrow to see if there's any negative impact in other places unless you otherwise have @benknoble

benknoble:
@dsifford I have not, my only due-diligence here has been in comments.

benknoble (approved):

benknoble:
@adriantrunzo thanks, that looks great

dsifford:
Sorry for the delay here. Just checked this out a bit more and LGTM for now. Thanks!

PR #256, opened 2021-07-22
Use termguicolors

lastra-dev:
Fixes neovim displaying cterm 256 colors, it should always use gui hexadecimal colors instead. this only applies if termguicolors is available for the terminal. this can save a lot of "colors not displaying correctly" issues.

Before:

After:

(difference in the background and status line)

### DIFF
--- a/autoload/dracula.vim
+++ b/autoload/dracula.vim
@@ -54,4 +54,13 @@ func! dracula#should_abort(...)
     return 0
 endfunction
 
+" Use termguicolors instead of 256
+if (has('nvim'))
+  let $NVIM_TUI_ENABLE_TRUE_COLOR = 1
+endif
+
+if (has('termguicolors'))
+  set termguicolors
+endif
+
 " vim: fdm=marker ts=2 sts=2 sw=2 fdl=0:


autoload/dracula.vim
```diff
@@ -54,4 +54,13 @@ func! dracula#should_abort(...)
     return 0
 endfunction
 
+" Use termguicolors instead of 256
+if (has('nvim'))
+  let $NVIM_TUI_ENABLE_TRUE_COLOR = 1
+endif
+
+if (has('termguicolors'))
```
benknoble, line 62: This is just wrong--vim supports the option, but not every terminal supports truecolor. It is an option we should not set for the user! They know better than we if this is what they want. We provide both GUI and terminal colors--the scheme is designed to work for both (though not perfectly as you point out).

autoload/dracula.vim
```diff
@@ -54,4 +54,13 @@ func! dracula#should_abort(...)
     return 0
 endfunction
 
+" Use termguicolors instead of 256
+if (has('nvim'))
+  let $NVIM_TUI_ENABLE_TRUE_COLOR = 1
```
benknoble, line 59: I have no idea what this does--is it neovim's version of `termguicolors`?

Does neovim ubiquitously support truecolor, even when running in terminals without it?

benknoble (commented):
Fundamentally I can't approve this. Not every terminal supports truecolor, and (while I happen to like the real colors) perhaps some are fine with the 256 color palette. This is not something we can or should force on the user, and I would consider it a breaking change.

autoload/dracula.vim
dsifford: FYI:

image

https://twitter.com/justinmk/status/730614264659120128

autoload/dracula.vim
dsifford: Agree. I don't believe we should be responsible for this.

autoload/dracula.vim
benknoble: Ok, so that would need fixed anyhow.

lastra-dev:
Ok, termguicolor should be enabled by the user if desired, maybe you can mention it at draculatheme.com/vim? Thank you :smile:

dsifford:
Appreciate the PR nevertheless @lastra-dev. Please feel free to open up more if you run into other issues or have other enhancements in mind.

PR #260, opened 2021-09-02
A better contrast for the tabline background

SlyBouhafs:
Here are two screenshots, the first one is the new setting and the second one is the old setting.
Screen Shot 2021-09-02 at 00 47 57
Screen Shot 2021-09-02 at 00 17 41

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -232,7 +232,7 @@ hi! link Question     DraculaFgBold
 hi! link Search       DraculaSearch
 call s:h('SignColumn', s:comment)
 hi! link TabLine      DraculaBoundary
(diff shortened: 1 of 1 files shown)


benknoble:
I don't have any issue with the change, but it looks like there's a spurious commit there: We don't need the "merge PR commit," just the change.

SlyBouhafs:
@benknoble Well, i'm not sure how to remove that merge PR commit, should i just do another PR?

dsifford:
@benknoble if you're cool with the change just do a squash merge and that stray commit won't matter.

benknoble:
Hm, I think I'll just cherry-pick it myself tomorrow. Don't worry about it @SlyBouhafs , I'll get it fixed up.

PR #270, opened 2021-12-03
fix(neovim): diagnostic highlights in v0.6.0

saviocmc:
Release notes:
https://github.com/neovim/neovim/releases/tag/v0.6.0

Breaking change commit:
https://github.com/neovim/neovim/pull/15585/commits/a5bbb932f9094098bd656d3f6be3c58344576709

Updated documentation:
https://neovim.io/doc/user/diagnostic.html#diagnostic-highlights

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -262,14 +262,14 @@ if has('nvim')
   hi! link LspReferenceText DraculaSelection
   hi! link LspReferenceRead DraculaSelection
   hi! link LspReferenceWrite DraculaSelection
-  hi! link LspDiagnosticsDefaultInformation DraculaCyan
-  hi! link LspDiagnosticsDefaultHint DraculaCyan
-  hi! link LspDiagnosticsDefaultError DraculaError
-  hi! link LspDiagnosticsDefaultWarning DraculaOrange
-  hi! link LspDiagnosticsUnderlineError DraculaErrorLine
-  hi! link LspDiagnosticsUnderlineHint DraculaInfoLine
-  hi! link LspDiagnosticsUnderlineInformation DraculaInfoLine
-  hi! link LspDiagnosticsUnderlineWarning DraculaWarnLine
+  hi! link DiagnosticInfo DraculaCyan
+  hi! link DiagnosticHint DraculaCyan
+  hi! link DiagnosticError DraculaError
+  hi! link DiagnosticWarn DraculaOrange
+  hi! link DiagnosticUnderlineError DraculaErrorLine
+  hi! link DiagnosticUnderlineHint DraculaInfoLine
+  hi! link DiagnosticUnderlineInfo DraculaInfoLine
+  hi! link DiagnosticUnderlineWarn DraculaWarnLine
 else
   hi! link SpecialKey DraculaPink
 endif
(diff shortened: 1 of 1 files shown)


benknoble:
Hm. Shouldn't we support both old versions and new versions? Maybe a conditional on the neovim version? (Or just eagerly create links to both sets?)

@dsifford what do you think? I don't think I want a plugin upgrade to break for folks still on neovim 0.5.x.

dsifford:
@benknoble Agree, we should just set both the old and the new at least for now.

saviocmc:
I agree we should not introduce breaking changes. At first when I saw the commit that introduced the change in neovim I thought they would keep links to the old names for compatibility, as stated [here in the commit](https://github.com/neovim/neovim/blob/a5bbb932f9094098bd656d3f6be3c58344576709/runtime/plugin/diagnostic.vim#L28). But now, I can't find the links anywhere in the neovim codebase, so I assume now they removed the links after the release. My bad.
So should I now just add links to the old names now?

benknoble:
> So should I now just add links to the old names now?

Yeah, let's do that.

colors/dracula.vim
```diff
@@ -262,14 +262,26 @@ if has('nvim')
...
+  " for backward compatibility with neovim v0.5.x
+  hi! link LspDiagnosticsDefaultInformation DiagnosticInfo
+  hi! link LspDiagnosticsDefaultHint DiagnosticHint
+  hi! link LspDiagnosticsDefaultError DiagnosticError
+  hi! link LspDiagnosticsDefaultWarning DiagnosticWarn
+  hi! link LspDiagnosticsUnderlineError DiagnosticUnderlineError
+  hi! link LspDiagnosticsUnderlineHint DiagnosticUnderlineHint
+  hi! link LspDiagnosticsUnderlineInformation DiagnosticUnderlineInfo
+  hi! link LspDiagnosticsUnderlineWarning DiagnosticUnderlineWarn
+
```
benknoble, line 284: If you reorganize, the diff will be simpler (_i.e._, put the old ones back first, where they were, with your comment; then have the new ones. linebreak before else not needed.)

```suggestion
  " Link old 'LspDiagnosticsDefault*' hl groups
  " for backward compatibility with neovim v0.5.x
  hi! link LspDiagnosticsDefaultInformation DiagnosticInfo
  hi! link LspDiagnosticsDefaultHint DiagnosticHint
  hi! link LspDiagnosticsDefaultError DiagnosticError
  hi! link LspDiagnosticsDefaultWarning DiagnosticWarn
  hi! link LspDiagnosticsUnderlineError DiagnosticUnderlineError
  hi! link LspDiagnosticsUnderlineHint DiagnosticUnderlineHint
  hi! link LspDiagnosticsUnderlineInformation DiagnosticUnderlineInfo
  hi! link LspDiagnosticsUnderlineWarning DiagnosticUnderlineWarn
  
  hi! link DiagnosticInfo DraculaCyan
  hi! link DiagnosticHint DraculaCyan
  hi! link DiagnosticError DraculaError
  hi! link DiagnosticWarn DraculaOrange
  hi! link DiagnosticUnderlineError DraculaErrorLine
  hi! link DiagnosticUnderlineHint DraculaInfoLine
  hi! link DiagnosticUnderlineInfo DraculaInfoLine
  hi! link DiagnosticUnderlineWarn DraculaWarnLine
```

benknoble:
Ok, looks reasonable to me. @dsifford have you tested this on neovim? If so I think we're ready to merge it.

dsifford:
@benknoble I have not had the opportunity to test, but I think this should be fine to merge in. We can address any issues if they pop up after but I think we're fairly safe with this.

PR #274, opened 2022-01-17
Add nvim-cmp menu highlight

yuys13:
Hi, thanks for maintaining a great color scheme!

When I was updating `CmpItemAbbrMatch`, I took a look at the other color schemes as well.
I noticed then that in tokyonight and kanagawa, `CmpItemMenu` is an inconspicuous color.
`CmpItemMenu` is `[LSP]` in the following image.
menu_before

CmpItemMenu is a completion source, and I consider it better to have an inconspicuous color, as shown in comments.
The image with this modification is shown below.
menu_after

What do you think about this modification?
Does it keep the Doracula feel?

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -133,6 +133,8 @@ if exists('g:loaded_cmp')
   hi! link CmpItemKindEvent DraculaFg
   hi! link CmpItemKindOperator DraculaPink
   hi! link CmpItemKindTypeParameter DraculaCyan
+
+  hi! link CmpItemMenu Comment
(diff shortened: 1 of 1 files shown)


dsifford (commented):
Looks fine to me. Thoughts @stnley @benknoble?

stnley:
LGTM. I think it makes sense the completion source be comment rather than FG.

dsifford:
That's enough for me! :shipit:

yuys13:
Thanks! I was wondering if I should change the color, but I'm glad you like it.

PR #275, opened 2022-01-18
Make dracula work with diffchar

ferdinandyb:
Screenshots and background in this issue:
https://github.com/rickhowe/diffchar.vim/issues/25

Code is by @rickhowe, not me.

Thanks for considering!

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -207,15 +207,16 @@ call s:h('StatusLineTermNC', s:none, s:bglight)
 call s:h('WildMenu', s:bg, s:purple, [s:attrs.bold])
 call s:h('CursorLine', s:none, s:subtle)
 
+call s:h('DiffAdd', s:green)
+call s:h('DiffChange', s:orange, s:none)
+call s:h('DiffText', s:bg, s:orange)
+call s:h('DiffDelete', s:red, s:bgdark)
+
 hi! link ColorColumn  DraculaBgDark
 hi! link CursorColumn CursorLine
 hi! link CursorLineNr DraculaYellow
-hi! link DiffAdd      DraculaGreen
 hi! link DiffAdded    DiffAdd
-hi! link DiffChange   DraculaDiffChange
-hi! link DiffDelete   DraculaDiffDelete
 hi! link DiffRemoved  DiffDelete
-hi! link DiffText     DraculaDiffText
 hi! link Directory    DraculaPurpleBold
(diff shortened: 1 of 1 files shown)


colors/dracula.vim
```diff
@@ -207,15 +207,16 @@ call s:h('StatusLineTermNC', s:none, s:bglight)
...
+call s:h('DiffDelete', s:red, s:bgdark)
+
 hi! link ColorColumn  DraculaBgDark
 hi! link CursorColumn CursorLine
 hi! link CursorLineNr DraculaYellow
-hi! link DiffAdd      DraculaGreen
 hi! link DiffAdded    DiffAdd
-hi! link DiffChange   DraculaDiffChange
-hi! link DiffDelete   DraculaDiffDelete
 hi! link DiffRemoved  DiffDelete
```
dsifford, line 219: What about `DiffAdded` and `DiffRemoved`? Do those not need to be highlighted this way?

dsifford (commented):
Thanks for the PR. Just a quick question, but otherwise LGTM

colors/dracula.vim
ferdinandyb: I don't really know the internals, so all I can say, that in practice they do not :) For the sake of symmetry they could be, but I don't really know if there are any downsides you'd need to consider.

dsifford: Hmm yeah, can you just make those the same as the other diff ones just for future-proofing?

ferdinandyb: should be done

dsifford (approved):
LGTM. 

Any last min thoughts, @benknoble?

benknoble:
I think this is really a temporary workaround, right? The real issue is that the other plugin didn't handle linked highlight groups correctly.

IIRC from threads on the mailing list, some folks are working on improving the vimscript API for handling highlights. Perhaps the plugin author needs to check that work and fix the plugin?

OTOH, I'm not really opposed to un-linking. Just want to make sure we're not taking on extra work for an out-of-band or out-of-scope issue.

benknoble:
Is this still something that we need? Did the upstream issue get resolved, and either way is this something we should consider merging?

ferdinandyb:
sorry for the late answer, tbh, I'm not sure what the situation is, I'll try to get around to checking this this week

ferdinandyb:
The diffchar plugin sorted this on their side, everything works as expected from dracula and diffchar master branches so this PR is now obsolete. Thanks!

PR #283, opened 2022-03-28
fix: support lazy-loading nvim-cmp

stnley:
Often users of [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) lazy-load this plugin. If users do that, then the highlight groups defined by this scheme don't get loaded with the current guard setup. 

This proposes an autocmd which hooks into the [documented](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt#L355)  `CmpReady` event, adding the highlights at that point.

Not sure how you feel about adding autocmds to the theme. I opted for a singular group `___dracula_cmp___` instead of a plugin-wide group (for example `___dracula___`).

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -102,7 +102,7 @@ endif
 " nvim-cmp: {{{
 " A completion engine plugin for neovim written in Lua.
 " https://github.com/hrsh7th/nvim-cmp
-if exists('g:loaded_cmp')
+function! s:set_cmp()
   hi! link CmpItemAbbrDeprecated DraculaError
 
   hi! link CmpItemAbbrMatch DraculaCyan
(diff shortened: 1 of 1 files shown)


dsifford:
@stnley Couldn't you just re-load `after/plugin/dracula.vim` with an autocmd on your end when you load the completion plugin?

stnley:
@dsifford Yeah that method words - if that is preferred than shipping with them I can close this, and keep in that in mind in the future!

benknoble:
Wouldn't this change mean that if nvim-cmp is _not_ lazy loaded, and happens to run before `after/plugin/dracula` (thus firing the event before the autocommand is installed) that things still wouldn't work?

I'm in favor of keeping our setup a little simpler :)

stnley:
@benknoble In my testing it did work as expected when lazy-loading and normal loading. However it seems like the consensus is to keep the setup as-is and direct user to implement their own autocmd when lazy-loading. That makes sense so I'll go ahead and close this. 

Thank you both for looking into it!

Issue #130, opened 2019-03-20
Diffs hard to read with cursorline over text


bdielman:
### What happened

When diffing files in Vim, changed sections are hard to run under the cursorline.

### What I expected to happen

In other colorschemes, the diff line supersedes the cursorline which makes the text easier to read (example provided with Gruvbox colorscheme).

### Screenshot

![Diff Image](https://user-images.githubusercontent.com/94547/54691802-461b8980-4afa-11e9-9171-dcde45da2b52.jpg)

### Machine Info

- **Vim type (`vim`/`gvim`/`neovim`)**: MacVim and using Vim in the terminal
- **Vim version**: 8.1.950
- **OS**: MacOS 10.14.3
- **Terminal/Terminal Emulator/VTE**: Terminal.app

benknoble:
That's bizarre... @dsifford @bdielman any ideas what's causing the difference?

dsifford:
I don't use cursorline personally but I can check into this later.

dsifford:
Just looked into this and I can't reproduce this. Is this still happening for you @bdielman?

bdielman:
@dsifford - I can but I'm not sure what could be causing it. I'll try later with the other plugins disabled to see if I can track it down.

bdielman:
Still happening even with my other plugins disabled. Since it's not reproducible I'm cool with closing the issue. 😄

PR #285, opened 2022-05-14
Fix Diffs hard to read with cursorline over text

sprloth:
Fix the problem Diffs hard to read with cursorline over text, https://github.com/dracula/vim/issues/130


benknoble:
Looks like a reasonable change, but it will be a while before I can test it. Do you have a screenshot or two?

sprloth:
before
屏幕截图 2022-05-23 204305

after
屏幕截图 2022-05-23 204554

PR #289, opened 2022-08-11
Neovim: Update tree-sitter support for CSS

tientrinh21:
Hello!
I found that [dracula/vim](https://github.com/dracula/vim) worked well with HTML & JS via tree-sitter but not CSS. So I took some time to add some highlights for it; I tried to make it look like Vscode as much as possible.

### What I linked:
*  Make `,`, `.`, `;`, etc. become pink: `TSPunctDelimiter` -> `DraculaPink` 
*  Make selectors become pink:  `TSType` -> `DraculaPink`
*  Make classes, properties become cyan: `TSProperty` -> `DraculaCyan`
*  Make URLs become orange: `TSError` -> `DraculaOrange` (I was inspired by how [shaunsingh/nord.nvim](https://github.com/shaunsingh/nord.nvim) did it)

### Before:
image

### After:
image

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -101,7 +101,6 @@ if exists('g:loaded_nvim_treesitter')
 	hi! link TSPunctDelimiter DraculaPink
 	hi! link TSType DraculaPink
 	hi! link TSProperty DraculaCyan
-	hi! link TSFunction DraculaGreenItalic
 	hi! link TSError DraculaOrange
 endif
 " }}}
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -97,6 +97,12 @@ if exists('g:loaded_nvim_treesitter')
   " HTML and JSX tag attributes. By default, this group is linked to TSProperty,
   " which in turn links to Identifer (white).
   hi! link TSTagAttribute DraculaGreenItalic
+	" CSS
+	hi! link TSPunctDelimiter DraculaPink
+	hi! link TSType DraculaPink
+	hi! link TSProperty DraculaCyan
+	hi! link TSFunction DraculaGreenItalic
+	hi! link TSError DraculaOrange
 endif
 " }}}
 " nvim-cmp: {{{


after/plugin/dracula.vim
```diff
@@ -97,6 +97,11 @@ if exists('g:loaded_nvim_treesitter')
   " HTML and JSX tag attributes. By default, this group is linked to TSProperty,
   " which in turn links to Identifer (white).
   hi! link TSTagAttribute DraculaGreenItalic
+	" CSS
+	hi! link TSPunctDelimiter DraculaPink
+	hi! link TSType DraculaPink
+	hi! link TSProperty DraculaCyan
+	hi! link TSError DraculaOrange
```
dsifford, line 104: I'm not sure this is going to be the proper way to go about this. `TSType`, for example, is the token used for all `type`s, in ever language. That for the most part should be `DraculaCyan`..

The way we need to go about this is be specific to these changes _just_ for css languages. There's a way to target those languages specifically, but you'll have to look into the docs for the tree sitter plugin to find that out because I can't recall off the top of my head how to do that.

dsifford (changes requested):

after/plugin/dracula.vim
benknoble: Seconded

NFrid: yea, you might make some queries for that.

For example, create `queries/css/highlights.scm` in the plugin's directory like this:
```scm
; inherits: css

[
 "#"
 ","
 "."
 ":"
 "::"
 ";"
 ] @keyword

(class_name) @type
(property_name) @type
(nesting_selector) @keyword
```

This will partially do the thing you did w/out linking any highlight groups 👽
image

p.s. you might find helpful `:TSPlaygroundToggle`, `:TSNodeUnderCursor` and `TSHighlightCapturesUnderCursor` commands of the `nvim-treesitter/playground` addon on your way.

ryouiki-tenkai:
someone working on this as there is stilll no css highlighting in neovim?

benknoble:
Neither Derek nor I is working on it, I believe. We requested some changes which have not been implemented.

Issue #276, opened 2022-02-01
the diff mode color


etorth:
![image](https://user-images.githubusercontent.com/1754214/151909904-e437d19e-243b-474a-a20b-75395621aa2f.png)

Everything looks good but the diff mode *may* need more contrast.
Not only the font color, if we can change the background color it's better.

benknoble:
Can you clarify where you want to see more contrast?

If we change the background color everywhere it's hard to identify word diffs (the background highlight in your screenshot shows where only part of a line changed, which is nice).

I suppose we could consider changing the background in a different way for the non-word diffs. Suggestions?

etorth:
![image](https://user-images.githubusercontent.com/1754214/152014464-7caf3fa1-baea-4d3f-be4f-918d5ac81700.png)

I mean something like this, although this one has a bad palette.
The diff lines needs *high-contrast background color* to those unchanged lines.
For people working on file diffs this is very helpful, because we just (need to) ignore all unchanged lines.
A partially changed line can still be highlighted, as in this example.

In this example I hate the write font in bright purple which make things worse.

benknoble:
I'm happy to add such a feature, but I don't have the time (or the eyeball, probably) to come up with the right colors. 

If you look in `:help dracula`, there are some useful instructions for customizing the theme (you could also edit the sources). Perhaps you could try some different colors and find an approach that works for you? Then we can discuss a concrete proposal and refine it. 

@dsifford if we added a high-contrast diff coloring, would you want that behind a feature flag or to just become the diff coloring? I can see how it would be useful, but I'm not 100% sure about changing the default.

dsifford:
I think in order to add this without a flag we'd need to have a bunch of people asking for it.

I'd say, to start off, it might be a good idea for @etorth to get a setup that they like working locally in their own config and then pitch the idea here with actual suggested changes in a PR that can be iterated on.

PR #292, opened 2022-10-27
support high contrast diff color

etorth:
fix of #276

with this change:

image

without this change:

image

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -97,6 +97,10 @@ if !exists('g:dracula_colorterm')
   let g:dracula_colorterm = 1
 endif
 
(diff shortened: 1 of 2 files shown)


etorth:
using red/green for high contrast.
Maybe not a good one, please suggest if you feel it worth merge and you have better combination.


image

colors/dracula.vim
```diff
@@ -97,6 +97,10 @@ if !exists('g:dracula_colorterm')
   let g:dracula_colorterm = 1
 endif
 
+if !exists('g:dracula_high_contrast_diff')
+  let g:dracula_high_contrast_diff = 0
```
benknoble, line 101: Please document this

PR #293, opened 2022-10-28
do not use green as high-contrast-diff bg color

etorth:
...since it hides selected text in changed lines.

Sorry first issue comes quickly.
As following, the ```tracebitmap``` in line 950 get hidden because selected text also uses green bg-color.

I created the ```s:blue``` since it's not used for now.
It should avoid any potential conflict with current all used color-configs.

without fix:

with fix:

### DIFF
--- a/autoload/dracula.vim
+++ b/autoload/dracula.vim
@@ -20,6 +20,7 @@ let g:dracula#palette.pink      = ['#FF79C6', 212]
 let g:dracula#palette.purple    = ['#BD93F9', 141]
 let g:dracula#palette.red       = ['#FF5555', 203]
 let g:dracula#palette.yellow    = ['#F1FA8C', 228]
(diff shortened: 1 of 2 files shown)


benknoble:
Oh darn, I didn't think about that.

Unfortunately introducing a _new_ color is probably not an option, unless it convincingly fits in to the existing palette. More at https://spec.draculatheme.com.

---

Can you explain what you mean by "selected text"? I'm wondering if there's another way to avoid this.

etorth:
> Oh darn, I didn't think about that.
> 
> Unfortunately introducing a _new_ color is probably not an option, unless it convincingly fits in to the existing palette. More at https://spec.draculatheme.com.
> 
> Can you explain what you mean by "selected text"? I'm wondering if there's another way to avoid this.

ok let me try to find a color from the existing palette.

PR #294, opened 2022-10-28
use yellow/purple as high-contrast-diff color

etorth:
continuation of #293 

use yellow/purple for high-contrast-diff color.
Looks purple/yellow is not used for text in buffer, should be safe now.

without fix:

wtih fix


benknoble:
Neat! I'll merge this and if we have to revisit later we can do that.

PR #299, opened 2022-12-01
Fix don't interrupt the cursorline

### LINKED ISSUE #295: Janky Cursorline Highlighting With Treesitter
### What happened
When using `set cursorline` in tandem with Treesitter, the Cursorline highlighting looks janky.
### What I expected to happen
The Cursorline would appear uniform across the whole line.
### Screenshot
what it looks like
image
what it should look like
image

### Machine Info

- **Vim type**: `neovim (GUI Client)`
- **Vim version**:  `nvim v0.80`
- **OS**: Windows

### Additional Info

This problem does get fixed when setting the colorscheme again.
The second image I provided is the same line but in visual line mode.

Note that visual mode and cursorline highlights are very similar in color. This makes it difficult for the user to determine whether or not they are in visual line selection with cursorline enabled. Perhaps make the visual mode highlight color darker?

saccarosium:
Fix #295
 
*before*
Screenshot 2022-12-01 at 20 03 20

*after*
Screenshot 2022-12-01 at 20 01 07

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -103,7 +103,7 @@ if exists('g:loaded_nvim_treesitter')
   if has('nvim-0.8.1')
     " # Misc
     hi! link @punctuation.delimiter Delimiter
-    hi! link @punctuation.bracket Normal
+    hi! link @punctuation.bracket DraculaFg
     hi! link @punctuation.special Special
     " # Constants
     hi! link @constant Constant
@@ -121,7 +121,7 @@ if exists('g:loaded_nvim_treesitter')
     hi! link @parameter DraculaOrangeItalic
     hi! link @parameter.reference DraculaOrange
     hi! link @field DraculaOrange
-    hi! link @property Normal
+    hi! link @property DraculaFg
     hi! link @constructor DraculaCyan
     " # Keywords
     hi! link @label DraculaPurpleItalic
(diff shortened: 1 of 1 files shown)


benknoble:
Hiya, sorry I've let this sit so long.

The screens look good, but can you explain why the fixes work? How certain are we they won't impact other highlights? For example, while I don't use NeoVim or Treesitter, [I sometimes prefer to dim brackets dramatically](https://github.com/benknoble/Dotfiles/blob/34443c53e5931eeef1ccdd547da787e8ec461e44/links/vim/after/syntax/racket.vim#L10). (To be clear, I don't think there are any issues, but I'm not 100% convinced yet.)

saccarosium:
The problem, same as in #295, is that because in some treesitter highilight groups were linked to `Normal` caused the cursorline to be "interrupted". The reason why is that Vim by default, if you don't assign anything to the bg highilght property, will make it something like `hi <highlight> fg=<color> bg=NONE`. But by linking an highilght to `Normal` you are overwriting that NONE with the `Normal` backgroud (which is the colorscheme background). 
The fix make it so instead of linking to Normal is linking to DraculaFg, which preserve the NONE background, making so the cursorline is not interrupted.

> How certain are we they won't impact other highlights? For example, while I don't use NeoVim or Treesitter, [I sometimes prefer to dim brackets dramatically](https://github.com/benknoble/Dotfiles/blob/34443c53e5931eeef1ccdd547da787e8ec461e44/links/vim/after/syntax/racket.vim#L10).

The fix will not affect any other highlights because I simply substitute `Normal` with `DraculaFg`, which are exatcly the same exept that `DraculaFg` has a bg=NONE.

Hope I explained the fix.

PR #304, opened 2023-04-28
Introduce gray color for SpecialKey

sh1r4s3:
I use UTF-8 chars for my custom `listchars` for `tabs` and `trail`. For example, in C code with tabs indents, `DraculaPink` is used for both `SpecialKey` and C keywords so it gets hard to read the code. I decided to switch `SpecialKey` color to something like gray color with a bit less blue component in it to distinguish it (it's not perfect though) from the comments color. Maybe this change might be useful.

The default scheme:
pink

With gray color:
gray

### DIFF
--- a/autoload/dracula.vim
+++ b/autoload/dracula.vim
@@ -14,6 +14,7 @@ let g:dracula#palette.selection = ['#44475A', 239]
 let g:dracula#palette.subtle    = ['#424450', 238]
 
 let g:dracula#palette.cyan      = ['#8BE9FD', 117]
+let g:dracula#palette.gray      = ['#616151', 128]
 let g:dracula#palette.green     = ['#50FA7B',  84]
 let g:dracula#palette.orange    = ['#FFB86C', 215]
 let g:dracula#palette.pink      = ['#FF79C6', 212]
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -44,6 +44,7 @@ let s:selection = g:dracula#palette.selection
 let s:subtle    = g:dracula#palette.subtle
 
 let s:cyan      = g:dracula#palette.cyan
+let s:gray      = g:dracula#palette.gray
 let s:green     = g:dracula#palette.green
 let s:orange    = g:dracula#palette.orange
 let s:pink      = g:dracula#palette.pink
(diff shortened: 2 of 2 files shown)


benknoble:
Some notes:
- We avoid colors outside of the dracula palette where possible to maintain consistency with other editor themes.
- `SpecialKey` affects more than just these `listchars` options (though some of them are `NonText`). NeoVim actually does slightly better here, using a dedicated highlight group for `listchars`-style things and leaving `SpecialKey` for, well, `SpecialKey`. Of note: `map` and `digraph` output, key-notation in `map` commands in `filetype=vim`, etc. Dimming these makes them less useable in most situations, at the cost of bright trailing spaces.
- It's entirely possible to keep these customizations for yourself if we don't include them as-is (which seems unlikely). See `:help dracula-customization` for the details on setting up a `ColorScheme` autocommand to override `DraculaSpecial` (or just `SpecialKey`) as needed.

sh1r4s3:
@benknoble thanks for the comment! I've also had a thought to simply change the color to `DraculaComment`. Also I can change `SpecialKey` to `Whitespace` for neovim. But I'm not sure if this PR is needed at all :) If not I can close it.

benknoble:
Yes, I would use `DraculaSubtle` or similar, personally. [Here's the reminder for NeoVim](https://github.com/dracula/vim/blob/eb577d47b0cfc9191bf04c414b4042d5f1a980f8/colors/dracula.vim#L270).

Frankly, the best thing to do is to convince Bram to add something similar for Vim so we can separate `SpecialKey` from "trailing spaces in `listchars`." In the meantime, I don't think we would accept this PR as-is.

Thanks for reminding us that we're interested in this, too, though!

sh1r4s3:
> Frankly, the best thing to do is to convince Bram to add something similar for Vim so we can separate `SpecialKey` from "trailing spaces in `listchars`.

Yes, it would be nice to have it one day!

I'll close this PR since this could be just a personal customization and no solution would be  objective :) However, I noticed that for neovim `SpecialKey` is `DraculaRed` whereas for vim it is `DraculaPink`. IMO `DraculaRed` looks is a bit better in the case mentioned in the first comment (attached a screenshot). I'm just wondering, is there any reason to differ `SpecialKey` for vim and neovim? AFAIU neovim uses `Whitespace` to highlight `listchars` and it is not gets set in the repo.
red

benknoble:
This was the commit that made it red (https://github.com/dracula/vim/commit/72a0d0d584e2b64dd7cd44047688d470e6e39f8f). You can also see that `SpecialKey` used to be `DraculaSubtle`. It's used in so many places that we made it Pink to stand out more without suggesting "error" in things like `:map` and `:digraph` notation.

PR #306, opened 2023-05-25
Disable background highlighting of `DraculaDiffDelete` when appropriate

noah-friedman:
Disable the background highlight for the `DraculaDiffDelete` group when:
- `g:dracula_colorterm` is false, AND
- `has('gui_running')` is false

Before:
Screenshot 2023-05-25 at 11 26 25 AM

After:
Screenshot 2023-05-25 at 11 27 48 AM

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -57,17 +57,17 @@ if has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
   hi! link GitSignsChangeLn DiffChange
   hi! link GitSignsChangeNr DiffChange
 
-	" Manually set `GitSignsDelete` when in a terminal where `g:dracula_colorterm`
-	" is false to remain consistent with the other GitSigns highlights.
-	if g:dracula_colorterm || has('gui_running') 
-		hi! link GitSignsDelete DiffDelete
-	else
-		hi! link GitSignsDelete DraculaRed
-	endif
-	" GitSignsDeleteLn and GitSignsDeleteNr should always match
-	" whatever GitSignsDelete is set to.
-	hi! link GitSignsDeleteLn	GitSignsDelete
-	hi! link GitSignsDeleteNr	GitSignsDelete
+  " Manually set `GitSignsDelete` when in a terminal where `g:dracula_colorterm`
+  " is false to remain consistent with the other GitSigns highlights.
+  if g:dracula_colorterm || has('gui_running') 
+    hi! link GitSignsDelete DiffDelete
+  else
+    hi! link GitSignsDelete DraculaRed
+  endif
+  " GitSignsDeleteLn and GitSignsDeleteNr should always match
+  " whatever GitSignsDelete is set to.
+  hi! link GitSignsDeleteLn GitSignsDelete
+  hi! link GitSignsDeleteNr GitSignsDelete
 endif
 " }}}
 " Tree-sitter: {{{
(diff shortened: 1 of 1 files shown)


noah-friedman:
That's a fair point. I'm more-so thinking about the background on the symbol in the signs column, but that's something I can take up with the maintainers of that plugin. I'll go ahead and close this PR.

after/plugin/dracula.vim
```diff
@@ -56,9 +56,18 @@ if has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
...
+	" is false to remain consistent with the other GitSigns highlights.
+	if g:dracula_colorterm || has('gui_running') 
+		hi! link GitSignsDelete DiffDelete
+	else
+		hi! link GitSignsDelete DraculaRed
+	endif
+	" GitSignsDeleteLn and GitSignsDeleteNr should always match
+	" whatever GitSignsDelete is set to.
+	hi! link GitSignsDeleteLn	GitSignsDelete
+	hi! link GitSignsDeleteNr	GitSignsDelete
```
benknoble, line 70: This indent looks wrong. Indent inside the `if`, sure, but no extra around it. Also double-check the indent amount (I can't recall what we've been using but it's probably 2 or 4 spaces).

benknoble:
Can you change the PR to just drop the reverted commit? No need for the noise in the history. (You can do this with an interactive rebase, for example.)

after/plugin/dracula.vim
noah-friedman: > This indent looks wrong. Indent inside the `if`, sure, but no extra around it. Also double-check the indent amount (I can't recall what we've been using but it's probably 2 or 4 spaces).

Done (was using tab literals instead of expanding. whoops).

noah-friedman:
> Can you change the PR to just drop the reverted commit? No need for the noise in the history. (You can do this with an interactive rebase, for example.)

Dropped. Here are new before and after screenshots (changed the Terminal background to make the change more visible):

Before:
Screenshot 2023-05-25 at 3 05 42 PM

After:
Screenshot 2023-05-25 at 3 07 19 PM

after/plugin/dracula.vim
benknoble, line 70:
```suggestion
  " Manually set `GitSignsDelete` when in a terminal where `g:dracula_colorterm`
  " is false to remain consistent with the other GitSigns highlights.
  if g:dracula_colorterm || has('gui_running') 
    hi! link GitSignsDelete DiffDelete
  else
    hi! link GitSignsDelete DraculaRed
  endif
  " GitSignsDeleteLn and GitSignsDeleteNr should always match
  " whatever GitSignsDelete is set to.
  hi! link GitSignsDeleteLn GitSignsDelete
  hi! link GitSignsDeleteNr GitSignsDelete
```
This is the indent I'm referring to (the "leading" indent). Your comment says you understood that, but I'm not sure that got picked into this PR.

benknoble:
The visual changes LGTM. Still dealing with code indent.

@dsifford ?

dsifford:
I'm cool with it if you are @benknoble (pending fixes to formatting)

after/plugin/dracula.vim
noah-friedman: I did change it, seems like it got axed when I squashed the commits together. Thanks for catching it!

noah-friedman:
Formatting issues resolved. Sorry about the indent, didn't realize it got cut out when I squashed the commits.

PR #307, opened 2023-06-11
Update GitSignsDelete highlight group to always link to DraculaRed

speelbarrow:
This is an add-on to the changes made in #306. Upon further usage I noticed that the gutter column used by the `GitSigns` plugin never uses the darker background that the `GitDiffDelete` highlight group uses. 

`GitSignsDelete` (new in #306) was set to link to `DraculaRed` _unless_ the user was in a GUI or had `g:colorterm` set true. However, in both of those cases the correct background to use still isn't that provided by `GitDiffDelete`.

I've updated the `GitSignsDelete` highlight group to **always** link to `DraculaRed` to keep it consistent with the background of the other signs as well as the column as a whole.

### DIFF
--- a/after/plugin/dracula.vim
+++ b/after/plugin/dracula.vim
@@ -57,17 +57,9 @@ if has('nvim-0.5') && luaeval("pcall(require, 'gitsigns')")
   hi! link GitSignsChangeLn DiffChange
   hi! link GitSignsChangeNr DiffChange
 
-  " Manually set `GitSignsDelete` when in a terminal where `g:dracula_colorterm`
-  " is false to remain consistent with the other GitSigns highlights.
-  if g:dracula_colorterm || has('gui_running') 
-    hi! link GitSignsDelete DiffDelete
-  else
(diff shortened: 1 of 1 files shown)


benknoble (approved):
I'm not sure I totally understand the shift between earlier and now, but I'm ok with the changes if it seems to be working and won't break anything for other folks.

dsifford:
Agree. Screenshot showing the differences would also be helpful for my understanding here.

speelbarrow:
Before (when `g:dracula_colorterm=1` or `has(“gui”)`):
image

After (in **all** contexts):
image

If you look at the images from #306 you’ll see that the latter image here matches the background of the `GitSigns` symbol when a line has been added/modified.

benknoble:
I'm good to merge, I think. I'll wait another few days to give Derek a chance to weigh in, and I'll take silence for tacit approval.

dsifford:
LGTM!

@benknoble I trust your judgement. Feel free to merge anything in the future that you feel is warranted!

PR #311, opened 2023-08-12
Add DraculaInlayHint and link from LspInlayHint

cmcfarlen:
The default display of the new LspInlayHint highlight group is difficult to read.  I added a new Dracula group and linked it to LspInlayHint.  I think this is pretty easy on the eyes and is similar to what the Dracula theme does on CLion.

Before:
Screenshot from 2023-08-12 16-27-26

After:
Screenshot from 2023-08-12 16-26-55

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -210,7 +210,7 @@ else
 endif
 
 call s:h('DraculaDiffText', s:bg, s:orange)
-call s:h('DraculaInlayHint', s:comment, s:bgdark, [s:attrs.italic])
+call s:h('DraculaInlayHint', s:comment, s:bgdark)
 
 " }}}2
 
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -210,6 +210,7 @@ else
 endif
 
 call s:h('DraculaDiffText', s:bg, s:orange)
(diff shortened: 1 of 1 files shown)


benknoble:
Can you share the default so that I can experiment a little?

@dsifford I'm assuming we don't have spec for this. But I think using a comment color is reasonable.

cmcfarlen:
The default for LspInlayHint links to NonText.  This is only in neovim-0.10 nightly builds.

Is this what you were asking for?

benknoble:
Yeah, I think so. I'll have to check if we have a NonText (we should). Just want to make sure I understand the changes :)

dsifford:
Seems reasonable, though I would caution using italics because some folks (particularly the visually impaired) tend to not like it. Would be better to just avoid if we could.

cmcfarlen:
Does this PR look ok now or are there other issues? It won't hurt my feelings if you want to tackle this another way later :smile:

benknoble:
I haven't had and won't have anytime soon the time to double-check what I want to. That said, if @dsifford approves, feel free to merge without me.

dsifford:
Looks like a reasonable change to me. Let's go for it.

Thanks @cmcfarlen!

PR #318, opened 2024-02-04
Add strikethrough attribute support

stnley:
This adds another configurable attribute for strikethrough text, along with a default foreground group.

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -77,6 +77,10 @@ if !exists('g:dracula_italic')
   let g:dracula_italic = 1
 endif
 
+if !exists('g:dracula_strikethrough')
+  let g:dracula_strikethrough = 1
+endif
+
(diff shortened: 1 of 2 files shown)


benknoble:
I don't see this used anywhere, so why do we need it? Perhaps you accidentally left off the use?

stnley:
I am preparing a PR with a treesitter highlight which uses this. I hoped it would be easier to process this separately. 

Admittedly it is just one group that will use it so if we don't want to have to maintain this no big deal.

benknoble:
Oh, that makes sense. Reading is easier broken up, but understanding the purpose helps.

benknoble:
I think I'd still like to see the follow-up to get a better idea of how this is used, esp. since I'd prefer not to merge this without a concrete use.

stnley:
Thanks for your patience, I've opened the follow up which uses this feature.

PR #321, opened 2024-04-01
Remove the use of the /after directory for defining plugin and syntax highligts.

### LINKED ISSUE #280: Highlight works from the start with treesitter, but is lost after sourcing vimrc.
Greetings,
I added Dracula to NVim and noticed that, when I source my vimrc the syntax highlighting changes.
I'm still a new user, so constantly changing vimrc every now and then.
Ended up with a minimal but the issue is still there, also read quite a bunch of solutions to no avail.
The ideal would be to have the same structure as VSCode (is this possible, today? even if it's not done atm), but the main issue here is making it not lose/change the highlights after sourcing.
Is there anything that can be done?
Thanks a lot!

Example js and html

```
call plug#begin()
Plug 'dracula/vim', {'as': 'dracula'}
Plug 'nvim-treesitter/nvim-treesitter', {'do': ':TSUpdate'}
call plug#end()

lua << EOF
    require'nvim-treesitter.configs'.setup {
        ensure_installed = "all",
        highlight = {
            enable = true,
            additional_vim_regex_highlighting = false,
        },
        indent = {
            enable = true
        },
    }
EOF

if !exists('g:syntax_on')
    syntax enable
endif
if has('termguicolors')
    set termguicolors
endif
let g:dracula_colorterm = 1
colorscheme dracula
```

- **Vim type (`vim`/`gvim`/`neovim`)**: neovim 
- **Vim version**: v0.7.0-dev+1149-g8bf3a3e30
- **OS**: Windows 10
- **Terminal/Terminal Emulator/VTE**: Windows Terminal / PowerShell
- **`TERM` environment variable**: vtpcon

### LINKED ISSUE #320: Cannot customize plugin specific highlight groups linked in /after files.
### What happened

I would like to overwrite the highlight group for GitGutterDelete that is linked in /after/plugin/dracula.vim: https://github.com/dracula/vim/blob/9fa89296884e47bbadc49ad959e37b9d1c24cafb/after/plugin/dracula.vim#L48

Note I just want to change GitGutterDelete, not DiffDelete.

Following the docs, I put my customization in a ColorScheme autocommand like so (near the top of my vimrc, before calling `colorscheme dracula`):

```vim
augroup Config
  autocmd!
augroup end

autocmd Config ColorScheme dracula call s:handle_colorscheme_dracula()

function! s:handle_colorscheme_dracula() abort
  highlight link CurSearch IncSearch
  highlight! link GitGutterDelete DraculaRed
endfunction
```

No matter what I try (with or without the bang !), the highlight group linked in the dracula plugin takes precedence:

```
:verbose hi GitGutterDelete                                                                                                                                                         
GitGutterDelete xxx links to DiffDelete                                                                                                                                             
        Last set from ~/.vim/pack/minpac/start/dracula/after/plugin/dracula.vim line 48
```

What am I missing? I assume the issue is that the dracula plugin puts files in `/after`, which is loaded after the vimrc, and uses `hi! link`. 

To reproduce my setup:

1. Create the following minimal vimrc file. In my demo I have called it minimal.vim:

```
augroup Config
  autocmd!
augroup end

autocmd Config ColorScheme dracula call s:handle_colorscheme_dracula()

function! s:handle_colorscheme_dracula() abort
  highlight link CurSearch IncSearch
  highlight! link GitGutterDelete DraculaRed
endfunction

filetype plugin indent on
syntax enable
set termguicolors
colorscheme dracula
```

1. Put the dracula theme files into the vim runtime path, for example ~/.vim/pack/plugins/start/dracula.
1. Open vim with vim -N -u ./minimal.vim.
1. `:verbose hi GitGutterDelete`

### What I expected to happen

I expect the output of `:verbose hi GitGutterDelete` to show `GitGutterDelete xxx links to DraculaRed`.

### Screenshot

Screenshot 2024-03-12 at 9 53 02 PM

### Machine Info

- **Vim type (`vim`/`gvim`/`neovim`)**: vim
- **Vim version**: 9.1.150
- **OS**: macOS 14.3.1
- **Terminal/Terminal Emulator/VTE**: Kitty 0.33.0
- **`TERM` environment variable**: xterm-kitty

adriantrunzo:
# Overview

Fixes #320 
Fixes #280 

Following the discussion in #320, this pull request is an attempt to remove the use of the `/after` directory for loading plugin and filetype colors. The goals of these changes are:

- To make the colorscheme easier to customize
- To allow for writing a re-loadable vimrc

# Visual Regression Checklist

## Plugins

- [ ] FZF
- [ ] ALE
- [x] CtrlP
- [ ] GitGutter
- [ ] GitSigns (Neovim)
- [ ] Treesitter (Neovim)
- [ ] Cmp (Neovim)

## Syntax

- [x] CSS
- [ ] Git Commit
- [x] HTML
- [x] Javascript/JSX
- [x] JSON
- [ ] Lua
- [ ] Markdown
- [ ] OCaml
- [ ] Perl
- [ ] PHP
- [ ] Python
- [ ] RST
- [ ] Ruby
- [ ] Rust
- [ ] Sass
- [ ] Shell
- [ ] Tex
- [ ] Typescript/TSX
- [ ] Vim
- [ ] XML
- [ ] Yaml

# Screenshots

For the following screenshots, I put the relevant plugins into my `~/.vim/pack/dracula/start` directory and used the following vimrc:

```
set nocompatible
filetype plugin indent on
syntax enable
set termguicolors
colorscheme dracula
```

#### Before:
Screenshot 2024-03-31 at 6 46 12 PM
#### After
Screenshot 2024-03-31 at 6 51 05 PM

#### Before
Screenshot 2024-03-31 at 7 17 29 PM
#### After
Screenshot 2024-03-31 at 7 26 13 PM

#### Before
Screenshot 2024-03-31 at 7 31 52 PM
#### After
Screenshot 2024-03-31 at 7 26 45 PM

#### Before
Screenshot 2024-03-31 at 7 22 44 PM
#### After
Screenshot 2024-03-31 at 7 29 34 PM

#### Before
Screenshot 2024-03-31 at 7 23 27 PM
#### After
Screenshot 2024-03-31 at 7 27 02 PM


benknoble:
Also, massive kudos for organizing such a large undertaking. Let us know if/where you need help.

benknoble:
First 2 commits should be squashed together so that the code moves from after to colors in a single step. Looks like the right idea, though I'm going to have to check if the bang in `hi!` is actually needed.

Since the highlight commands are now in colors, what happens if I open a file and do `:edit!` to "reload" the file? Do the dracula links get washed away or...? (I'm thinking about the ordering, here: Dracula runs early, presumably, because it's a colorscheme. But syntax files run when I edit a file, so wouldn't they establish their links/colors over top of the Dracula ones? From your screens it appears to be working but I confess I've gotten myself confused 😅)

colors/dracula.vim
```diff
@@ -373,6 +373,628 @@ hi! link helpCommand DraculaPurple
...
+  hi! link CmpItemKindEnumMember DraculaPurple
+  hi! link CmpItemKindConstant DraculaPurple
+  hi! link CmpItemKindStruct DraculaPink
+  hi! link CmpItemKindEvent DraculaFg
+  hi! link CmpItemKindOperator DraculaPink
+  hi! link CmpItemKindTypeParameter DraculaCyan
+
+  hi! link CmpItemMenu Comment
+  " }}}
+
```
benknoble, line 981: When I check this out locally, it looks like there might be some trailing spaces here.

benknoble:
I made some tweaks to my local config (mostly due to the removal of `dracula#should_abort`), but I haven't noticed any issues with this branch so far.

colors/dracula.vim
adriantrunzo: Removed, thanks.

adriantrunzo:
> First 2 commits should be squashed together so that the code moves from after to colors in a single step.

I've rebased.

> Since the highlight commands are now in colors, what happens if I open a file and do :edit! to "reload" the file? Do the dracula links get washed away or...? (I'm thinking about the ordering, here: Dracula runs early, presumably, because it's a colorscheme. But syntax files run when I edit a file, so wouldn't they establish their links/colors over top of the Dracula ones? From your screens it appears to be working but I confess I've gotten myself confused 😅)

From my research, all Vim syntax files use `hi def` (`:hi-default`) to define default highlight links that will never overwrite an already defined highlight link ([example](https://github.com/vim/vim/blob/3c4d2e7a09134b72b129d26db374a8a8325c2779/runtime/syntax/python.vim#L300), [example](https://github.com/vim/vim/blob/3c4d2e7a09134b72b129d26db374a8a8325c2779/runtime/syntax/lua.vim#L410)).

> Looks like the right idea, though I'm going to have to check if the bang in hi! is actually needed.

Good question. All of the other colorschemes I've looked at use `hi! link` for defining links, or redefine highlight groups outright with `hi {group} ...` ([example](https://github.com/nordtheme/vim/blob/main/colors/nord.vim), [example](https://github.com/sainnhe/everforest/blob/master/autoload/everforest.vim)). Even the builtin colorschemes use `hi! link` ([example](https://github.com/vim/vim/blob/master/runtime/colors/desert.vim)). I wonder if removing the bang will create a bunch of issues from folks who are linking highlight groups for whatever reason before setting `:colorscheme dracula` in their vimrc, as Dracula would not overwrite those links and the colorscheme would look broken.

All of the official colorscheme customization methods listed in `:color-scheme` are compatible with `hi! link` as they would be called after the colorscheme is loaded and the user can use their own bang. We can specifically call out the need to use the bang in the docs. For example, my vimrc has a block like this (the bang is necessary):

```vim
autocmd Config ColorScheme dracula call s:handle_colorscheme_dracula()

function! s:handle_colorscheme_dracula() abort
  highlight! link GitGutterDelete DraculaRed
endfunction
```

adriantrunzo:
I've updated the customization part of the documentation to mention the need for the bang (!) when changing highlight group links created by the colorscheme.

adriantrunzo:
> I made some tweaks to my local config (mostly due to the removal of dracula#should_abort)

Did you have to make those tweaks because this pull request created a breaking change for you, or were you simply removing code that is no longer necessary?

benknoble:
> > I made some tweaks to my local config (mostly due to the removal of dracula#should_abort)
> 
> Did you have to make those tweaks because this pull request created a breaking change for you, or were you simply removing code that is no longer necessary?

Both, but my configuration was sort of broken anyway. Moving the guarded code to the proper `autocmd` and getting rid of references to `dracula#should_abort` did the trick.

> I've updated the customization part of the documentation to mention the need for the bang (!) when changing highlight group links created by the colorscheme.

Great, thanks!

RE: `:highlight`, for now let's keep doing what we're doing. It's working :)

benknoble:
I haven't noticed any real issues here—I know there are plugins left to test, but I'm hoping to be able to merge this soon (partly because I'd like to also get it merged with the Pro theme before the Alucard release). Is there anything we can do to help? I'm probably more concerned with plugins than filetypes at the moment.

I use ALE and haven't seen any problems. We can tick that off, I think.

I'm also happy to address specific problems later, at this point.

adriantrunzo:
@benknoble My apologies. I have been trying to find a way to automate the screenshots because they are time consuming.

I use FZF, GitGutter, and Coc (which has no special treatment in dracula at the moment) and haven't noticed any issues.

> I'm also happy to address specific problems later, at this point.

Thanks! I will be available too. Again, I was trying to find a way to make all of screenshots. I also don't use Neovim, but I don't see any obvious reasons for regressions there.

adriantrunzo:
I've marked this as ready for review so it is mergeable. I've set aside time this weekend to make more screenshots, but I can also focus on just testing and reporting the results.

benknoble:
Excellent. @dsifford any thoughts? I'll aim to merge Sunday or Monday if I don't hear back.

dsifford:
Sorry for the radio silence. LGTM 🫡

PR #62, opened 2018-03-23
Massive theme overhaul

dsifford:
This PR is an attempt to normalize the Vim Dracula implementation in its current form according to the running [specification](https://dsifford.github.io/dracula-spec/) that I [proposed](https://github.com/dracula/dracula-theme/issues/232) in the main repository.

The PR is a baseline for further scrutinization. It is in no way complete, however I did have the opportunity to test and confirm all the languages that were explicitly targeted in the existing implementation are highlighted appropriately in accordance to the above specification.

Along with the obvious fact that this is more or less a complete rewrite, the following notible changes were made:

- Several dozen custom `Dracula<xxx>` highlight scopes were made, which existing highlight scopes are then linked to to allow for slightly better startup performance and, more importantly, easier debugging with `:hi<CR>`
- This introduces several user customizations that allows users to disable text style attributes if they don't want them or their terminals don't support them (underline, italic, bold, undercurl, inverse).

Anyway, I'll leave it at that for now. Curious to hear what you all think.

Closes #25
Closes #37
Closes #45
Closes #46
Closes #48
Closes #50

Also, this negates the need for #42, so that can probably be closed.

CC: @zenorocha, @benknoble

PR #336, opened 2026-03-06
feat: support mini.statusline plugin

### LINKED ISSUE #335: Can we support mini.statusline?
I see we've implemented support for mini.icons.
Could we also support the mini.statusline from the same series?

adriantrunzo:
Fixes #335 

I've tried to replicate the airline theme for the mode colors and adhere to the defaults of the mini.statusline plugin for the rest. Airline doesn't support a separate highlight group for command mode, so I've chosen the cyan color.

Documentation for the highlight groups: https://github.com/nvim-mini/mini.nvim/blob/cad365c212fb1e332cb93fa8f72697125799d00a/doc/mini-statusline.txt#L48

The screenshots below are using the default configuration for the mini.statusline plugin.

---

Screenshot 2026-03-14 at 11 58 20 AM

Screenshot 2026-03-14 at 11 58 28 AM

Screenshot 2026-03-14 at 11 59 13 AM

Screenshot 2026-03-14 at 11 58 40 AM

Screenshot 2026-03-14 at 11 58 55 AM

Screenshot 2026-03-14 at 11 59 36 AM

Screenshot 2026-03-14 at 12 04 46 PM

Screenshot 2026-03-14 at 12 10 51 PM

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -170,30 +170,35 @@ call s:h('DraculaSubtle', s:subtle)
 
 call s:h('DraculaCyan', s:cyan)
 call s:h('DraculaCyanItalic', s:cyan, s:none, [s:attrs.italic])
+call s:h('DraculaCyanInverse', s:cyan, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaGreen', s:green)
 call s:h('DraculaGreenBold', s:green, s:none, [s:attrs.bold])
 call s:h('DraculaGreenItalic', s:green, s:none, [s:attrs.italic])
 call s:h('DraculaGreenItalicUnderline', s:green, s:none, [s:attrs.italic, s:attrs.underline])
+call s:h('DraculaGreenInverse', s:green, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaOrange', s:orange)
 call s:h('DraculaOrangeBold', s:orange, s:none, [s:attrs.bold])
 call s:h('DraculaOrangeItalic', s:orange, s:none, [s:attrs.italic])
 call s:h('DraculaOrangeBoldItalic', s:orange, s:none, [s:attrs.bold, s:attrs.italic])
-call s:h('DraculaOrangeInverse', s:bg, s:orange)
+call s:h('DraculaOrangeInverse', s:orange, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaPink', s:pink)
 call s:h('DraculaPinkItalic', s:pink, s:none, [s:attrs.italic])
+call s:h('DraculaPinkInverse', s:pink, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaPurple', s:purple)
 call s:h('DraculaPurpleBold', s:purple, s:none, [s:attrs.bold])
 call s:h('DraculaPurpleItalic', s:purple, s:none, [s:attrs.italic])
+call s:h('DraculaPurpleInverse', s:purple, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaRed', s:red)
-call s:h('DraculaRedInverse', s:fg, s:red)
+call s:h('DraculaRedInverse', s:red, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaYellow', s:yellow)
 call s:h('DraculaYellowItalic', s:yellow, s:none, [s:attrs.italic])
+call s:h('DraculaYellowInverse', s:yellow, s:bg, [s:attrs.inverse])
 
 call s:h('DraculaError', s:red, s:none, [], s:red)
 
@@ -202,7 +207,6 @@ call s:h('DraculaWarnLine', s:none, s:none, [s:attrs.undercurl], s:orange)
 call s:h('DraculaInfoLine', s:none, s:none, [s:attrs.undercurl], s:cyan)
 
 call s:h('DraculaTodo', s:cyan, s:none, [s:attrs.bold, s:attrs.inverse])
-call s:h('DraculaSearch', s:green, s:none, [s:attrs.inverse])
 call s:h('DraculaBoundary', s:comment, s:bgdark)
 call s:h('DraculaWinSeparator', s:comment, s:bgdark)
 call s:h('DraculaLink', s:cyan, s:none, [s:attrs.underline])
@@ -233,6 +237,9 @@ call s:h('StatusLineTerm', s:none, s:bglighter, [s:attrs.bold])
 call s:h('StatusLineTermNC', s:none, s:bglight)
 call s:h('WildMenu', s:bg, s:purple, [s:attrs.bold])
 call s:h('CursorLine', s:none, s:subtle)
+
+" Maintain the DraculaSearch group for backwards compatibility.
+hi! link DraculaSearch DraculaGreenInverse
 
 hi! link ColorColumn  DraculaBgDark
 hi! link CursorColumn CursorLine
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -1062,6 +1062,19 @@ if has('nvim')
   hi! link MiniIconsRed DraculaRed
   hi! link MiniIconsYellow DraculaYellow
   " }}}
+
+  " nvim-mini/mini.statusline {{{
+  hi! link MiniStatuslineDevInfo StatusLine
+  hi! link MiniStatuslineFileInfo StatusLine
+  hi! link MiniStatuslineFilename StatusLineNC
+  hi! link MiniStatuslineInactive StatusLineNC
+  hi! link MiniStatuslineModeCommand DraculaCyanInverse
+  hi! link MiniStatuslineModeInsert DraculaGreenInverse
+  hi! link MiniStatuslineModeNormal DraculaPurpleInverse
+  hi! link MiniStatuslineModeOther DraculaGreenInverse
+  hi! link MiniStatuslineModeReplace DraculaOrangeInverse
+  hi! link MiniStatuslineModeVisual DraculaYellowInverse
+  " }}}
 endif
 " }}}
 


adriantrunzo:
@cnjhb Feel free to try out this patch locally and let me know how it looks.

cnjhb:
> @cnjhb Feel free to try out this patch locally and let me know how it looks.

@adriantrunzo That's fine,
but wouldn't it be better to use different colors for command mode and normal mode?

cnjhb (approved):

colors/dracula.vim
```diff
@@ -1051,6 +1051,19 @@ if has('nvim')
   hi! link MiniIconsRed DraculaRed
   hi! link MiniIconsYellow DraculaYellow
   " }}}
+
+  " nvim-mini/mini.statusline {{{
+  call s:h("MiniStatuslineDevInfo", s:fg, s:bg)
+  call s:h("MiniStatuslineFileInfo", s:fg, s:comment)
+  call s:h("MiniStatuslineFilename", s:fg, s:selection)
+  call s:h("MiniStatuslineInactive", s:fg, s:selection)
+  call s:h("MiniStatuslineModeCommand", s:bg, s:purple)
```
cnjhb: I think the command mode should use cyan to better distinguish it from the normal mode's purple.

adriantrunzo:
> but wouldn't it be better to use different colors for command mode and normal mode?

Good question. It looks like the existing airline and lightline themes don't specify a separate color for command mode. Nor do I think airline supports a different color for command mode: https://github.com/vim-airline/vim-airline/blob/4ab7c731fe64672412fbe0723831928785c84931/autoload/airline/themes/dark.vim#L16.

Do you have any suggestions about which color to use? We haven't used Cyan or Pink, but they may be too bright? Red would look like an error.

To play it safe, we could just match airline and let people customize the `MiniStatuslineModeCommand` highlight group in their nvim configuration.

cnjhb:
> > but wouldn't it be better to use different colors for command mode and normal mode?
> 
> Good question. It looks like the existing airline and lightline themes don't specify a separate color for command mode. Nor do I think airline supports a different color for command mode: https://github.com/vim-airline/vim-airline/blob/4ab7c731fe64672412fbe0723831928785c84931/autoload/airline/themes/dark.vim#L16.
> 
> Do you have any suggestions about which color to use? We haven't used Cyan or Pink, but they may be too bright? Red would look like an error.
> 
> To play it safe, we could just match airline and let people customize the `MiniStatuslineModeCommand` highlight group in their nvim configuration.

It doesn't matter if it's brighter, because the text is black.
The colors in both visual mode and insert mode are quite bright and look quite suitable.

colors/dracula.vim
```diff
@@ -1051,6 +1051,19 @@ if has('nvim')
...
+  call s:h("MiniStatuslineModeInsert", s:bg, s:green)
+  call s:h("MiniStatuslineModeNormal", s:bg, s:purple)
+  call s:h("MiniStatuslineModeOther", s:bg, s:green)
+  call s:h("MiniStatuslineModeReplace", s:bg, s:orange)
+  call s:h("MiniStatuslineModeVisual", s:bg, s:yellow)
```
benknoble: > ```
>   call s:h("MiniStatuslineDevInfo", s:fg, s:bg)
> ```

I think this is "just" `DraculaNormal`, actually.

> ```
>   call s:h("MiniStatuslineFileInfo", s:fg, s:comment)
>   call s:h("MiniStatuslineFilename", s:fg, s:selection)
>   call s:h("MiniStatuslineInactive", s:fg, s:selection)
> ```

We have `s:bglighter`, `s:bglight` for things like this, too. See how `StatusLine` and `StatusLineNC` are defined.

> ```
>   call s:h("MiniStatuslineModeCommand", s:bg, s:purple)
>   call s:h("MiniStatuslineModeInsert", s:bg, s:green)
>   call s:h("MiniStatuslineModeNormal", s:bg, s:purple)
>   call s:h("MiniStatuslineModeOther", s:bg, s:green)
>   call s:h("MiniStatuslineModeReplace", s:bg, s:orange)
>   call s:h("MiniStatuslineModeVisual", s:bg, s:yellow)
> ```

We have a few "Inverse" colors, but they inconsistently put text in foreground or background colors. (We also inconsistently use the "inverse" attribute.) Perhaps as a preliminary step we should first cleanup these inconsistencies; then, introduce inverse colors for our palette; then, use them here?

adriantrunzo: > I think this is "just" DraculaNormal, actually.

Do you mean just `Normal`? I don't see any highlight group called `DraculaNormal`. https://github.com/dracula/vim/blob/6b0f9cca6263f60ef961d139541a999aeee97006/colors/dracula.vim#L229

> We have s:bglighter, s:bglight for things like this, too. See how StatusLine and StatusLineNC are defined.

Ok, I'll take a look at switching to these groups.

> We have a few "Inverse" colors, but they inconsistently put text in foreground or background colors.

I see exactly two: `DraculaOrangeInverse` and `DraculaRedInverse`. I assume the inconsistency you'd like to correct is the white text in `DraculaRedInverse`? Not sure if that white text was for readability reasons, but `DraculaRedInverse` is only used in one place, so this change shouldn't be too disruptive.

> Perhaps as a preliminary step we should first cleanup these inconsistencies; then, introduce inverse colors for our palette; then, use them here?

Sounds good, I'm happy to tackle this refactor.

benknoble: > > I think this is "just" DraculaNormal, actually.
> 
> Do you mean just `Normal`? I don't see any highlight group called `DraculaNormal`.

Yep, sorry.
 
> https://github.com/dracula/vim/blob/6b0f9cca6263f60ef961d139541a999aeee97006/colors/dracula.vim#L229
> 
> > We have s:bglighter, s:bglight for things like this, too. See how StatusLine and StatusLineNC are defined.
> 
> Ok, I'll take a look at switching to these groups.
> 
> > We have a few "Inverse" colors, but they inconsistently put text in foreground or background colors.
> 
> I see exactly two: `DraculaOrangeInverse` and `DraculaRedInverse`. I assume the inconsistency you'd like to correct is the white text in `DraculaRedInverse`? Not sure if that white text was for readability reasons, but `DraculaRedInverse` is only used in one place, so this change shouldn't be too disruptive.

Yeah. I don't know why it's white-on-red either, but I don't find it too readable now that I'm looking at it[^1]. It gets used for the standard `ErrorMsg`, but I think that might be more readable with `s:bg` as the text color (I find it so in a short test).

(There's also a "green inverse" but it might not have a name other than `DraculaSearch`.)

The other "inconsistency" is whether they use `s:attrs.inverse` or just swap fg/bg in `s:h()`. If you don't mess with that, it's fine, but it might be nice to clean up some day.

> > Perhaps as a preliminary step we should first cleanup these inconsistencies; then, introduce inverse colors for our palette; then, use them here?
> 
> Sounds good, I'm happy to tackle this refactor.

Much appreciated!

[^1]: This goes all the way back to that massive refactor we (@dsifford) did in #62; I haven't tried blaming within that PR for now.

dsifford: If y'all land on a combination that is more accessible, then I'm game to make the change. I trust your judgement @benknoble

adriantrunzo: I think black-on-red generally has more contrast than white-on-red. Here's a contrast comparison using the colors from dracula:

Screenshot 2026-03-11 at 4 42 18 PM

Screenshot 2026-03-11 at 4 44 21 PM

Screenshots from https://color.adobe.com/create/color-contrast-analyzer.

> (There's also a "green inverse" but it might not have a name other than DraculaSearch.)

Ah, thanks!

benknoble: Funny, I find white-on-red easier to read in those screenshots, but in Vim the opposite. I'd go with dark-on-red for consistency for now and we can always swap back later.

adriantrunzo: Sounds good, I'll have time to work more on this refactor this weekend.

adriantrunzo:
@benknoble I've pushed a new commit with the refactored inverse colors. A few points worth further discussion:
- I've explicitly set the inverse groups to use `s:bg` instead of `s:none` for the foreground color. This approach differs from `DraculaTodo`, for example. The reason is that with `s:none` the color underneath actually becomes the foreground (text) color. For the statusline (and potentially other cases), the color underneath can be a lighter color that reduces contrast and readability.
- I've removed the `DraculaSearch` highlight group because it's now redundant with `DraculaGreenInverse` and linked `Search` directly to `DraculaGreenInverse`. This change is technically a breaking change for folks that were referencing `DraculaSearch` in their personal configurations. Would you like me to add back a definition like `hi! link DraculaSearch DraculaGreenInverse` to maintain compatability?

@cnjhb I choose cyan for the command mode color. Please see the screenshots in the pull request description.

benknoble:
> @benknoble I've pushed a new commit with the refactored inverse colors. A few points worth further discussion:
> 
>     * I've explicitly set the inverse groups to use `s:bg` instead of `s:none` for the foreground color. This approach differs from `DraculaTodo`, for example. The reason is that with `s:none` the color underneath actually becomes the foreground (text) color. For the statusline (and potentially other cases), the color underneath can be a lighter color that reduces contrast and readability.

This is great, thanks.

Would you mind splitting into 2 patches?

1. Universally establish Inverse color groups
2. Support mini.statusline

If not, I can do it locally before merging.

>     * I've removed the `DraculaSearch` highlight group because it's now redundant with `DraculaGreenInverse` and linked `Search` directly to `DraculaGreenInverse`. This change is technically a breaking change for folks that were referencing `DraculaSearch` in their personal configurations. Would you like me to add back a definition like `hi! link DraculaSearch DraculaGreenInverse` to maintain compatability?

Yep, I'd prefer not to change (for now) the public interface by removing groups, so let's go ahead and keep `DraculaSearch` by whatever mechanism makes sense.

adriantrunzo:
> Would you mind splitting into 2 patches?
> 
>     1. Universally establish Inverse color groups
> 
>     2. Support mini.statusline
> 
> 
> If not, I can do it locally before merging.

Done. I've included both commits in this pull request, but also created a separate pull request with just the inverse group changes: https://github.com/dracula/vim/pull/338. I don't know what your git workflow is, but hopefully you have what you need.

> Yep, I'd prefer not to change (for now) the public interface by removing groups, so let's go ahead and keep DraculaSearch by whatever mechanism makes sense.

Sure, I've retained `DraculaSearch`.

benknoble:
> > Would you mind splitting into 2 patches?
> > ```
> > 1. Universally establish Inverse color groups
> > 
> > 2. Support mini.statusline
> > ```
> > If not, I can do it locally before merging.
> 
> Done. I've included both commits in this pull request, but also created a separate pull request with just the inverse group changes: #338. I don't know what your git workflow is, but hopefully you have what you need.

Thanks. I'm pretty comfortable with Git in weird ways, so the separate PR wasn't necessary.

> > Yep, I'd prefer not to change (for now) the public interface by removing groups, so let's go ahead and keep DraculaSearch by whatever mechanism makes sense.
> 
> Sure, I've retained `DraculaSearch`.

Yep, eyeballed the range-diff between the most recent versions and checked the tree diffs.

I made a few commit message tweaks (see range-diff below); merged as 84fbc42 (feat: support mini.statusline, 2026-03-15).

<details><summary>range-diff: add some more commit detail</summary>

```
1:  072313e ! 1:  1270100 feat: establish universal inverse color highlight groups
    @@ Metadata
      ## Commit message ##
         feat: establish universal inverse color highlight groups
     
    +    There are a number of colors missing inverse groups, and furthermore the
    +    existing ones are inconsistent (should they use inverse attributes or
    +    manually swap fg/bg? what should the text color be?). Since we are going
    +    to want more inverse colors to support a statusline plugin soon, go
    +    ahead and resolve these issues.
    +
    +    Keep DraculaSearch for compatibility, even though DraculaGreenInverse is
    +    a suitable replacement.
    +
      ## colors/dracula.vim ##
     @@ colors/dracula.vim: call s:h('DraculaSubtle', s:subtle)
      
2:  824c400 ! 2:  84fbc42 feat: support mini.statusline
    @@ Metadata
      ## Commit message ##
         feat: support mini.statusline
     
    +    Docs: https://github.com/nvim-mini/mini.nvim/blob/cad365c212fb1e332cb93fa8f72697125799d00a/doc/mini-statusline.txt#L48
    +    Fix: https://github.com/dracula/vim/issues/335
    +
      ## colors/dracula.vim ##
     @@ colors/dracula.vim: if has('nvim')
        hi! link MiniIconsRed DraculaRed
```

</details>

PR #337, opened 2026-03-14
feat: support added, changed, and removed highlight groups

adriantrunzo:
The Vim `syntax.txt` help document [lists three preferred highlight group names](https://vimhelp.org/syntax.txt.html#:~:text=%2AAdded%20added%20line%20in%20a%20diff%20%2AChanged%20changed%20line%20in%20a%20diff%20%2ARemoved%20removed%20line%20in%20a%20diff) that we don't currently support: `Added`, `Changed`, and `Removed`.

Since these groups are preferred names, some plugins (e.g. `mini.diff`) use them by default. Defining these groups in the dracula color scheme should improve out-of-the-box plugin support.

Screenshot 2026-03-14 at 2 46 48 PM

### DIFF
--- a/colors/dracula.vim
+++ b/colors/dracula.vim
@@ -331,6 +331,10 @@ hi! link Comment DraculaComment
 hi! link Underlined DraculaFgUnderline
 hi! link Todo DraculaTodo
 
+hi! link Added DiffAdded
+hi! link Changed DiffChange
+hi! link Removed DiffRemoved
+
 hi! link Error DraculaError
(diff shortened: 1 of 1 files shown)


benknoble:
Indeed. We do support `DiffAdd`/`DiffChange`/`DiffText`/`DiffDelete` (`:help :diff`), but it would be smart to support these as well.

Merged as d7563db (link more standard groups, 2026-03-14), thanks!

PR #338, opened 2026-03-15
feat: establish universal inverse color highlight groups

adriantrunzo:
Extracted from #336. These changes standardize highlight groups for inverse colors. Of note:
- The foreground (text) color for each inverse group is `s:bg` instead of `s:none` to ensure sufficient contrast. The `DraculaRedInverse` group used for error messages is no longer white text on a red background.
- The `DraculaSearch` is functionally redundant with `DraculaGreenInverse`, but retained as a semantic highlight group for backwards compatibility. `DraculaSearch` is now just a linked group.

Screenshot 2026-03-15 at 9 01 19 AM

Screenshot 2026-03-15 at 9 01 45 AM


benknoble:
Merged as 1270100 (feat: establish universal inverse color highlight groups, 2026-03-15).

PR #339, opened 2026-04-08
Fix comment with direct project page URL

clach04:


benknoble:
I think the closest thing to the original is https://github.com/dracula/dracula-theme, actually. Care to update it that way instead?

clach04:
Done :-)

benknoble:
Thanks Derek! Wild weekend :)

